PR #1, opened 2015-05-21
Removing composer.lock

gabrielsch:


samsonasik:
:+1:

weierophinney:
No. Shipping with a composer.lock file is considered a best practice at this time, and ensures all developers are testing against the same set of dependencies.

Maks3w:
Pro: All developers use the same dev dependencies

Con: Non-dev dependencies are locked and software is not tested with latest dependencies.

Proposal:
- Fix dev versions with "=" constraint in composer.json
- Add a test build in the matrix with composer --prefer-lowest

danizord:
@Maks3w :+1:

Ocramius:
> Shipping with a composer.lock file is considered a best practice at this time

This is a bad practice.

We are just making it harder to discover bugs caused by dependency updates, which in turn causes more issues when this package is used downstream by others (with a different dependency set).

I don't know where this "best practice" comes from, but it's not useful at all, and is actually harmful.

Re-opening.

Ocramius:
Note: this has already been beaten to death multiple times in other discussions (that I'm currently unable to find due to the amount of colliding search results).

`composer.lock` is contributor DX, but is terrible user DX.

Maks3w:
Probably Composer.lock was added to the repo due the BC violations from "semantic versioning" php-cs-fixer.

This has been resolved in components repositories removing the semantic versioning constraint "~1.7" with "1.7.*"

As I said there are better alternatives like lock version inside the composer.json

gianarb:
In my opinion `composer.lock` into the vcs is a "best practice" for an application.. Because I needed of a strong control of my version.. I can choose which dependency update, and I'm secure that others are fixed in a precise commit.

This is a library and in my opinion is a best practice remove this file because we have already decided the dependency's version in the `composer.json`

Maks3w:
Notice the point of @weierophinney is about lock developer tooling and not library dependencies.

It's reasonable all developers to test with the same tools

Ocramius:
> It's reasonable all developers to test with the same tools

Right, so it is "latest", not "frozen" ;-)

Additionally, `composer.lock` doesn't really apply when every dev environment is hugely different (people with PHP 5.4, 5.5, 5.6, 7.0-dev, etc), as dependency resolution for us may be hugely different from deps resolution of other people.

I'm repeating myself tho, I'll find the link to the dead-horse-beating-discussion.

Ocramius:
There we go: thephpleague/tactician#6

@ anyone else, please read through that before posting any further comments that may simply repeat the discussion

weierophinney:
Read the full thread, and agree (it was where I was leaning anyways, but having seen so many "always commit the `composer.lock`" slides and bullet points the past months, I was doubting myself). :+1:

Maks3w:
@gabrielsch For to consider this PR as complete you need to add composer.lock in .gitignore

If not the PR is not fixing future additions.

PR #7, opened 2015-06-07
Utils::getArity fails on callables in array format

tklever:
This patch allows Utils::getArity to correctly return the arity on callables passed in array format

i.e.

```php
array($this, 'method')
array('SomeClass', 'someMethod')
```

### DIFF
--- a/test/UtilsTest.php
+++ b/test/UtilsTest.php
@@ -13,6 +13,7 @@ use PHPUnit_Framework_TestCase as TestCase;
 use Zend\Stratigility\Dispatch;
 use Zend\Stratigility\MiddlewarePipe;
 use Zend\Stratigility\Utils;
+use ZendTest\Stratigility\TestAsset\NormalHandler;
 
 class UtilsTest extends TestCase
 {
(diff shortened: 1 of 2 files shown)


test/UtilsTest.php
```diff
@@ -24,6 +25,7 @@ public function callablesWithVaryingArity()
             }, 2],
             'invokable' => [new Dispatch(), 5],
             'interface' => [new MiddlewarePipe(), 2], // 2 REQUIRED arguments!
+            'callable'  => [array(new NormalHandler(), 'handle'), 3]
```
Maks3w: Please, use short array syntax

tklever: Fixed. Force of habit I suppose.

weierophinney:
Two notes:

- First, read the `CONTRIBUTING.md`; bugfixes should be submitted against the master branch, not the develop branch (that branch is reserved for new features). Right now it's not a big deal, as they're identical, but as the project grows, this will become increasingly important.
- Second, in reviewing, I came up with another case: `Class::method`. When I added that to the data providers, it failed; I've got code I'll commit with the merge that will resolve it, however. :)

Thanks for the patch, @tklever !

tklever:
@weierophinney No Problem, thank YOU for your continued hard work on all things Zend (and PSR7 and Middleware, etc., etc., etc.)

I pushed this against develop vs. master because I saw it as an enhancement (we're supporting an additional callback format) vs. a bug (we support this and it's broken). I can certainly see cases to be made for either direction, but I'm glad it was good enough to get merged.

PR #11, opened 2015-06-19
Update Utils.php

spiffyjr:
If $out/$next is nullable then `getNumberOfRequiredParameters()` will return 3 for a valid error handler with a null $out/$next causing the error detection to fail unless you use the ErrorMiddlewareInterface.

I think it's intended to use `getNumberOfParameters()` here.

**EDIT** There's a few other places in that file that would need changed but I'll wait until I hear back from you.

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -32,7 +32,7 @@ abstract class Utils
                 }
 
                 $r = new ReflectionMethod($callable, $method);
-                return $r->getNumberOfRequiredParameters();
+                return $r->getNumberOfParameters();
             }
             return 0;
         }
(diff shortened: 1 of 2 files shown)


spiffyjr:
I'll have to figure out the tests but wanted your (@weierophinney) input first.

weierophinney:
@spiffyjr Error handlers are defined as having *exactly* four *required* parameters; i.e., the `$out/$next` argument is not nullable.

Interestingly, I'm considering making the `$out/$next` argument on middleware *required*; having it nullable means that technically you need to check to see if it's non-empty before you call it, which I think is a bit of a burden for development.

spiffyjr:
Alright, that's fine. I assumed the final argument ($next/out) was not required.

I agree with the last argument being required. It's become a pain to type `return $next ? $next($req, $res) : $res;` continually.

PR #13, opened 2015-06-23
Extract `FinalHandler::getStatusCode` to Utils

weierophinney:
Extracted `FinalHandler::getStatusCode` to `Utils::getStatusCode`, making it a public static method (per @SpiffyJr). This allows re-use of that logic in custom final handlers.


spiffyjr:
:+1: - Perfect.

PR #17, opened 2015-07-19
Added ide generated files to gitignore.

HardieBoeve:


.gitignore
```diff
@@ -1,3 +1,13 @@
+.buildpath
+.DS_Store
+.idea
+.project
+.settings/
+.*.sw*
+.*.un~
+nbproject
+tmp/
+
```
samsonasik: remove the blank line

weierophinney:
Shouldn't these be in individual users' global gitignore rules, and not in the repository?

I have, in my `$HOME/.gitconfig`, the setting `core.excludesfile` set to the value of `$HOME/.gitignore-global`; that file contains exclusions for IDE-generated files. I recommend all contributors do the same.

Maks3w:
I don't remind to see a PR with foreign files added. I think the level of mature is enough so there is no need to exclude those files.

HardieBoeve:
Ok, closed as there are better alternatives to exclude these files.

PR #19, opened 2015-07-19
Bumped phpunit to latest stable release.

HardieBoeve:


Maks3w:
why?

weierophinney:
@Maks3w - we've done this for all the components, so, if nothing else, for consistency, and to allow using features from the newest versions (e.g. prophecy).

Maks3w:
@weierophinney If that is the intention (prophecy) then all components should bump to ^4.6

If the intention is consistency then ^4.0 should be the choice

PR #25, opened 2015-08-25
Next should clone $queue for long-lived scripts

weierophinney:
https://github.com/zendframework/zend-stratigility/blob/716c647fb76303c76c0c254bb4c5bc628aefa7e1/src/Next.php#L48

$queue should be cloned so that long-lived scripts can function. Without cloning the pipe will only ever work the first time and the queue will be empty on the next iteration.

### DIFF
--- a/src/Next.php
+++ b/src/Next.php
@@ -40,12 +40,16 @@ class Next
     private $removed = '';
 
     /**
+     * Constructor.
+     *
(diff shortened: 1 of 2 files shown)


weierophinney:
Patch created and issue converted to a pull request; please review, @spiffyjr . :)

spiffyjr:
Looks good and works in the use-case I was describing.

PR #30, opened 2015-09-10
Throw exception if response is already completed

### LINKED ISSUE #24: PSR-7 violations
I just spent two hours trying to find out why a middleware was unable to set a header, and it turns out [this](https://github.com/zendframework/zend-stratigility/blob/master/src/Http/Response.php#L147) line of code violates [this](https://github.com/php-fig/http-message/blob/master/src/MessageInterface.php#L124) clause of the spec:

> MUST return an instance that has the new and/or updated header and value.

The idea of a response being "complete" isn't even a thing in the spec.

Simply *ignoring* function calls in this manner makes it *very* hard to debug anything, and is extremely risky. If I called `setHeader()` (or any number of other methods) clearly I'm calling them for one reason, and one reason only - quietly ignoring this call is a clear violation of the spec.

Debugging this type of silent error is an absolute nightmare. Components and middleware worked perfectly in unit and integration tests, but then quietly failed when implemented in production code.

I'm baffled as to why you would think this feature is necessary in the first place? Request and Response models are already immutable - what possible reason could you have to disable an an already-immutable model instance from generating a new model?

If a given piece of middleware decides that a response is "complete", quite simply, it won't delegate control to the next middleware on the stack. Instead, this feature allows you to delegate control to the next middleware, but disabling it from actually doing anything - it will run, but can't actually *do* anything.

Am I crazy for thinking this is crazy??

codeliner:
- Closes #24

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -9,6 +9,7 @@
 
 namespace Zend\Stratigility\Http;
 
+use RuntimeException;
 use Psr\Http\Message\ResponseInterface as PsrResponseInterface;
 use Psr\Http\Message\StreamInterface;
 
@@ -53,14 +54,13 @@ class Response implements
     /**
      * Write data to the response body
      *
-     * Proxies to the underlying stream and writes the provided data to it.
-     *
-     * @param string $data
+     * {@inheritdoc}
+     * @throws RuntimeException if response is already completed
      */
     public function write($data)
     {
         if ($this->complete) {
-            return $this;
+            throw $this->responseIsAlreadyCompleted(__METHOD__);
         }
 
         $this->getBody()->write($data);
@@ -140,11 +140,12 @@ class Response implements
      * Proxy to PsrResponseInterface::withBody()
      *
      * {@inheritdoc}
+     * @throws RuntimeException if response is already completed
      */
     public function withBody(StreamInterface $body)
     {
         if ($this->complete) {
-            return $this;
+            throw $this->responseIsAlreadyCompleted(__METHOD__);
         }
 
         $new = $this->psrResponse->withBody($body);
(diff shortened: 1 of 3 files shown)


Ocramius:
@weierophinney can you let me know if this PR is legit?

I'm not experienced with the API, so I don't know if we're fixing broken API or actually changing it.

codeliner:
@Ocramius This change was discussed with @weierophinney in #24 
It is a BC break and should be scheduled for next major release.

weierophinney:
Next minor release; we're still in 0.*   :-)

codeliner:
@weierophinney [v1.1.0](https://github.com/zendframework/zend-stratigility/releases/tag/1.1.0) doesn't look like a dev release ;) Confused with zend-expressive? Or did I miss something?

weierophinney:
@codeliner Yep. That's what I get by answering notifications via email.

codeliner:
;)

src/Http/Response.php
```diff
@@ -270,4 +275,17 @@ public function getReasonPhrase()
...
+    /**
+     * @param string $detectedInMethod
+     * @return RuntimeException
+     */
+    private function responseIsAlreadyCompleted($detectedInMethod)
+    {
+        return new RuntimeException(sprintf(
+            'Calling %s::%s is not possible. Response is already completed',
+            __CLASS__,
+            $detectedInMethod
```
weierophinney: `__METHOD__` includes both the method name **and** the class name; as such, you can simplify `%s::%s` to just `%s`, and remove the `__CLASS__` argument.

weierophinney:
As noted in #24, and my yourself, this is a BC break, *but also fixes bad behavior*. As such, I'm going to schedule this for the next *minor* release, where we can call it out in the release notes.

PR #38, opened 2015-12-30
moved test in determinaton of border

pine3ree:
checking route identity to '/' with possible shortcircuit return before comparing string lengths (this test was always performed and winning if true anyway)

### DIFF
--- a/src/Next.php
+++ b/src/Next.php
@@ -164,8 +164,8 @@ class Next
         if ($route === '/') {
             return '/';
         }
-
-        return (strlen($path) > strlen($route)) ? $path[strlen($route)] : '';
+        $routeLength = strlen($route);
+        return (strlen($path) > $routeLength) ? $path[$routeLength] : '';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Next.php
```diff
@@ -161,11 +161,11 @@ private function resetPath(ServerRequestInterface $request)
...
-        $border = (strlen($path) > strlen($route))
-            ? $path[strlen($route)]
-            : '';
-        $border = ($route === '/') ? '/' : $border;
-        return $border;
+        if ($route === '/') {
+            return '/';
+        }
+
+        return (strlen($path) > strlen($route)) ? $path[strlen($route)] : '';
```
Maks3w: Could you extract length of $route to a local var? By this way we can avoid to do the operation two times

pine3ree: @Maks3w 
Hello, I usually do that and I have thought about that in this case, but sometimes I get flagged as micro-optimizator, so i didn't want to push it :-)

trivial matter a bout coding style:
I usually leave a blank line before `return`, but i noticed that mwof is not a big fan of it so I  removed it in the last commit

thanks and kind regards

PR #44, opened 2016-01-24
Minor optimization

bakura10:
This change prevents the usage of the slow call_user_func. The tests run on my machine but I didn't remember if this mechanism was not added on PHP 5.5 or 5.6... so I decided to do the PR and see if it passes on all PHP versions :).

### DIFF
--- a/src/Dispatch.php
+++ b/src/Dispatch.php
@@ -74,11 +74,11 @@ class Dispatch
 
         try {
             if ($hasError && $arity === 4) {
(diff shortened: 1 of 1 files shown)


bakura10:
I just realized that this package is supposed to be PHP 5.4 compatible, but Travis does not run on it. Expected @weierophinney ?

weierophinney:
Updated travis to run 5.4 tests as part of this (after first verifying tests ran locally under 5.4!)

Issue #45, opened 2016-01-29
FinalHandler should not display stack traces by default


michaelmoussa:
[This](https://github.com/zendframework/zend-stratigility/blob/1.1.2/src/FinalHandler.php#L117-L121) seems like a dangerous default to me, and it contradicts the Expressive documentation [here](https://github.com/zendframework/zend-expressive/blob/1.0.0/doc/book/features/error-handling.md#error-handling).

Specifically:

> If the FinalHandler was initialized with an option indicating that it is in development mode, it writes the exception stack trace to the response body.

However, the default initialization of the `FinalHandler` (which is [what Expressive does](https://github.com/zendframework/zend-expressive/blob/2db114f9c888e5ce06c808c44c47f70324be4d5c/src/ErrorMiddlewarePipe.php#L73) in the default error pipeline) leaves the `env` option unset, which causes `! isset($this->options['env'])` to evaluate to `true` and creates the Development error message. You'd actually _need_ to specify `['env' => 'production']` in the constructor in order to _not_ display the stack trace.

I think the reverse would be a safer default - you must explicitly state that you're in development in order for the stack trace to appear (which is what the Expressive documentation indicates).

Specifically: `if (isset($options['env']) && $options['env'] === 'development')`

Either that, or correct the Expressive documentation.

I like the former better, but I realize it's _technically_ a BC break (if your application needs to display stack traces in Production by default, that is...), so it should be discussed (assuming I'm not the only one who thinks this is an issue).

Thoughts?

weierophinney:
Definitely a bug. While it's a BC break, the defaults should always be production-safe, so I'd consider the current behavior a bug. Users can adapt if they need to have the stack trace displayed.

Send me a PR! :+1:

PR #46, opened 2016-02-02
Finalhandler production stacktrace fix

michaelmoussa:
@weierophinney this addresses #45 and makes it possible to create the `FinalHandler` in a factory and worry about the original response later.

One thing I wasn't sure about was whether error details should be in the case of `'env' !== 'production'` vs. `'env' === 'development'`. Production is surely "NO" and development is surely "YES", but other environments like staging, qa, etc. may depend on the implementor. Perhaps a boolean `display_error_details` option would be better than `env`? If so, what would you do with the original `env`? Remove it completely? Or automatically set `display_error_details` to `true` if `env != production` and `false` otherwise for BC?

### DIFF
--- a/src/FinalHandler.php
+++ b/src/FinalHandler.php
@@ -121,7 +121,8 @@ class FinalHandler
     private function handleError($error, RequestInterface $request, ResponseInterface $response)
     {
         $response = $response->withStatus(
-            Utils::getStatusCode($error, $response)
+            Utils::getStatusCode($error, $response),
+            $response->getReasonPhrase()
         );
 
         $message = $response->getReasonPhrase() ?: 'Unknown Error';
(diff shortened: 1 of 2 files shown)


michaelmoussa:
Addendum is a tiny bonus to fix another `FinalHandler` glitch.

nesl247:
Just ran into this issue myself and then I remembered you mentioning this PR. This is a huge issue in my opinion, so hopefully it gets merged in soon.

weierophinney:
I'd do `$env !== 'production'`, and the default will be production; that way you can override the value in local environments, but it defaults to "safe for production". I can make that change on merge.

weierophinney:
> I can make that change on merge.

Never mind; that's how it already reads. :smile:

weierophinney:
Marking as 1.2.0, as it introduces new functionality.

weierophinney:
Merged to develop for release with 1.2.0.

michaelmoussa:
:+1: thanks!!

PR #47, opened 2016-02-07
Added write JSON to Response decorator

arcostasi:
Write data to the response body with JSON encode and prepare to return an HTTP JSON response to the client. This method is not part of the PSR-7 standard, but it would be of great use in the zend-expressive for the API developers.

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -11,6 +11,7 @@ namespace Zend\Stratigility\Http;
 
 use Psr\Http\Message\ResponseInterface as PsrResponseInterface;
 use Psr\Http\Message\StreamInterface;
+use InvalidArgumentException;
 
 /**
  * Response decorator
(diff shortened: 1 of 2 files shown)


bakura10:
Hi,

This is not needed, you can already create a JsonResponse from Diactoros: https://github.com/zendframework/zend-diactoros/blob/master/src/Response/JsonResponse.php

arcostasi:
Hi Michaël Gallego, I know, but I would like to use the response passed by reference in the action methods. I don't like to get instantiated every time the new JsonResponse from Diactoros.
For an API developer it is very common to use JSON all the time.

Follow my exemple solution:

```
use Zend\Stratigility\Http\ResponseInterface;

public function ping(ServerRequestInterface $request, ResponseInterface $response, callable $next = null)
{
    $data = [
        'ack' => time()
    ];

    return $response->writeJson($data);
}
```
Follow my study zend-expressive project:
https://github.com/arcostasi/expressive-mvc-architecture

Best regards,

Anderson

weierophinney:
@arcostasi We have no plan to create implementation-specific methods in the base response class; the whole point of it is to be a faithful and robust PSR-7 implementation. Adding methods such as this means that we'd be introducing concretions that users may come to depend on, when they should be depending on the PSR-7 functionality.

One note:
> I would like to use the response passed by reference in the action methods. I don't like to get instantiated every time the new JsonResponse from Diactoros.

Whenever you call a `with*()` method, you're creating a new instance *regardless*, as these methods clone and return the new instance. So if your desire is because of not wanting to create a new instance, then it's misguided.

The concrete implementations in the `Zend\Diactoros\Response` namespace can be used for situations such as this. If you're going to have concrete functionality that diverges from the base PSR-7 functionality, these are what you should use. Otherwise, consume the PSR-7 functionality, or create helpers that you can use that do all the work at once:

```php
// Where "$jsonResponse()" is a function or callback that
// converts $data to JSON, injects it into the response, sets
// the content-type header, and the status:
$response = $jsonResponse($response, $data);
```

bakura10:
Also this could be a good use of trait ;).

arcostasi:
I understand, thanks Mr. Matthew O'Phinney and Mr. Michaël Gallego, I try implement an trait. :)

PR #37, opened 2015-12-24
Catch throwable for php7

fabiocarneiro:

PR #49, opened 2016-03-07
Hotfix - #37 - catch also php 7 throwables

Ocramius:
Provides tests for #37

### DIFF
--- a/test/DispatchTest.php
+++ b/test/DispatchTest.php
@@ -18,6 +18,16 @@ use Zend\Stratigility\Route;
 
 class DispatchTest extends TestCase
 {
+    /**
+     * @var \Zend\Stratigility\Http\Request|\PHPUnit_Framework_MockObject_MockObject
+     */
+    private $request;
+
+    /**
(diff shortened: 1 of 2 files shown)


nesl247:
Would this not fail on PHP 5.6 because the class `Throwable` doesn't exist?

msheakoski:
@nesl247 PHP < 7 doesn't appear to complain when Throwable is inside of a catch block. See example: http://sandbox.onlinephpfunctions.com/code/66997de5290d58dca15462e867224c9da62e3e14

nesl247:
Very interesting. I really hope this gets merged in soon then.

Ocramius:
@nesl247 type-hints are not causing any autoloading, so it would just be ignored ;-)

test/DispatchTest.php
```diff
@@ -199,4 +209,47 @@ public function testShouldAllowDispatchingPsr7Instances()
         $result = $dispatch($route, $err, $request->reveal(), $response->reveal(), $next);
         $this->assertSame($response->reveal(), $result);
     }
+
+    /**
+     * @requires PHP 7.0
```
weierophinney: That's nifty; didn't know about that annotation!

PR #56, opened 2016-04-27
Fixed broken unit test that was doing nothing

GeeH:


GeeH:
Sorry meant to make that on my own fork as I'm messing with CI server, my bad.

PR #67, opened 2016-09-27
Remove FinalHandler, ErrorMiddlewareInterface, and Dispatch

weierophinney:
This pull request is issued against the dev-2.0.0 branch, and builds directly on the changes in #66, with the following additional changes:

- Removes the `FinalHandler` class.
- Removes the `ErrorMiddlewareInterface`.
- Removes the `Dispatch` class; since all middleware is assumed to have the same signature now, and error handling can be done in middleware such as the `ErrorHandler`, it's redundant.
- Updates the `MiddlewareInterface` to make the third argument required, and rename it to `$next`.
  - Updates all implementations to follow suit.
- Updates `Next`:
  - Removes the `$done` constructor argument and property.
  - Removes the `$dispatch` property.
  - Inlines the code to dispatch middleware, as only one signature is now supported.
  - Returns the result of invoking middleware directly; it no longer verifies a response was returned. (This is no longer necessary, as it can be performed by classes such as the `ErrorHandler`.)
  - Returns the response provided at invocation directly if no more middleware is queued.
- Updates `MiddlewarePipe`:
  - Passes the result of invoking the next layer of the queue to the provided `$next` argument.
  - If the next layer of the queue does not return a response, the original response is passed to the `$next` argument.

All tests were audited to check for features no longer supported, and documentation has been updated.


weierophinney:
I've merged #66 to develop, and rebased dev-2.0.0 and this branch from there, as it is a continuation of the features started on that PR.

Ready for review! Pinging @zendframework/community-review-team @michaelmoussa and @xtreamwayz !

doc/book/api.md
```diff
@@ -112,6 +96,15 @@ function ($request, $response, $next)
 }
 ```
 
+> ### Do not pass an altered response
```
michaelmoussa: This threw me off a bit. The example demonstrated passing an altered response, but then we recommend _not_ doing that without showing how to do it the recommended way.

Perhaps, after "_or returning a brand new response entirely_" we add: 

> [...] as shown below:
>
> ```php
> function ($request, $response, $next)
> {
>     $nextResponse = $next($request, $response);
>     return $nextResponse->withAddedHeader('Cache-Control', [
>         'public',
>         'max-age=18600',
>         's-maxage=18600',
>     ]);
> }
> ```

?

michaelmoussa: **Note:** The ` $updated = $response->addHeader('Cache-Control', [` line under **Providing an altered response:** should be `->withAddedHeader`.

doc/book/api.md
```diff
@@ -163,56 +161,6 @@ And, if not calling `$next()`, returning the response instance:
 return $response;
 ```
 
-### Raising an error condition
```
michaelmoussa: I know error handling is covered elsewhere, but should we at least mention here that raising an error condition is as simple as throwing an exception, and then link to the documentation on error handling?

doc/book/api.md
xtreamwayz: `$response->addHeader` should be `$response->withAddedHeader` for all 3 examples under **Providing an altered response**, **Providing both an altered request and response** and **Returning a response to complete the request**. 

Since this code is also in 1.3, a PR to fix this might be in place.

xtreamwayz: Created PR #69 to address this and some more.

doc/book/api.md
weierophinney: Great feedback; I've re-instated this section, but now with an example demonstrating raising an exception from middleware.

doc/book/api.md
weierophinney: @michaelmoussa Thanks for the feedback! I've updated the examples to only demonstrate manipulating the returned response, instead of the response at invocation.

weierophinney: @xtreamwayz thanks for the PR; merged, and I've rebased this PR with the changes.

michaelmoussa:
👍 LGTM now @weierophinney. Sad to see `getArity(...)` go, though. Such a cool name! :)

weierophinney:
> Sad to see getArity(...) go

Me too, but as it's not used anymore, no need to keep it!

PR #70, opened 2016-09-29
Deprecate Http subnamespace

weierophinney:
Per the [RFC on removing the request and response decorators](https://gist.github.com/weierophinney/ecbd2ba572600a1bfc86840b4cfeca6b), this patch deprecates them for the upcoming 1.3 release.

The following changes were made:

Deprecated
----------

- `Zend\Stratigility\Http\Request`
  - The "psr" request is now injected with an additional attribute, `originalRequest`, allowing retrieval using standard PSR-7 methods.
  - Additionally, the methods `getCurrentRequest()` and `getOriginalRequest()` now emit deprecation notices.
- `Zend\Stratigility\Http\ResponseInterface`
  - Additionally, the methods `write()`, `end()`, `isComplete()`, and `getOriginalResponse()` now emit deprecation notices.
- `Zend\Stratigility\Http\Response`

Changed
-------

- `Zend\Stratigility\FinalHandler`
  - now pulls the original request via the `originalRequest` attribute instead.
  - no longer decorates the response, and instead writes to the PSR-7 response body directly (vs via the response `write()` method).

Added
-----

- `Zend\Stratigility\Middleware\OriginalMessages` is middleware that will inject the request instance passed to deeper layers with three attributes:
  - `originalRequest`, representing the request provided to it
  - `originalResponse`, representing the response provided to it
  - `originalUri`, representing the URI composed by the request provided to it
  We now recommend that if you need any of these artifacts, you pipe this middleware as the first layer of your application.

### DIFF
--- a/doc/book/migration/to-v2.md
+++ b/doc/book/migration/to-v2.md
@@ -8,6 +8,55 @@ To help you prepare your code for version 2, version 1.3.0 provides several
 forwards compatibility features to assist you in the process. However, some
 changes will still require changes to your code following the 2.0 release.
 
+## Original request, response, and URI
+
+In the original 1.X releases, Stratigility would decorate the request and
+response instances with `Zend\Stratigility\Http\Request` and
+`Zend\Stratigility\Http\Response`, respectively. This was done originally to
+facilitate access to the incoming request in cases of nested layers, where the
+URI path may have been truncated (`Next` truncates matched paths when executing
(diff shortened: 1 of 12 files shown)


weierophinney:
The next in the 1.3/2.0 series of deprecations and improvements. Pinging @zendframework/community-review-team , @michaelmoussa , and @xtreamwayz for review.

PR #71, opened 2016-09-29
Remove HTTP namespace

weierophinney:
This patch is against dev-2.0.0 and builds on #70, removing the classes and interfaces from the `Zend\Stratigility\Http` namespace deprecated by that patch.

Internally, this only affects `MiddlewarePipe` on this branch, as that class was decorating the incoming request and response at invocation; however, no other code was utilizing their features.

This patch targets the 2.0.0 release.


weierophinney:
Marking as a WIP until #70 is merged to develop.

weierophinney:
With #70 merged, ready for review; pinging @zendframework/community-review-team , @michaelmoussa , and @xtreamwayz.

weierophinney:
> I was wondering, the WhoopsErrorHandler checks for the Zend\Stratigility\Http\Request and grabs the original request from it. But I guess if you push it first in the pipeline it already has access to the original request, before any other middleware can mess with it.

Correct, and once we have a 1.3 version of Stratigility released, I'll update that handler to use the `originalRequest` attribute instead.

michaelmoussa:
👍  LGTM too.

xtreamwayz:
I really like all the changes. It makes understanding and customizing error handling a lot easier.

weierophinney:
With this in place, I've updated the branch-aliases across all branches to:

- dev-master => 1.2.x-dev
- dev-develop => 1.3.0-dev
- dev-dev-2.0.0 => 2.0.0-dev

As such, you should be able to test against the upcoming 1.3 or 2.0 versions using appropriate constraints (e.g., `^1.3.0@dev`, `^2.0.0@dev`).

PR #74, opened 2016-11-02
Remove aliases for PSR-7 interfaces from MiddlewareInterface

fabiocarneiro:
I propose to remove the `MiddlewareInterface` aliases for the PSR-7 interfaces. Its super annoying that most of the PSR-7 implementations (including diactoros) use `Response` as class name, then you have either to alias the Response implementation with something else or remove this aliases manually from your middleware (Considering they were automatically imported by your IDE when you implemented the `MiddlewareInterface`).

If you think about it, it adds no benefit, because in reality what is being expected there is a instance for the interface, so there is no benefit of hiding it in the arguments. On the other side, variable name makes sense, since you'll be receiving a concrete.

### DIFF
--- a/src/MiddlewareInterface.php
+++ b/src/MiddlewareInterface.php
@@ -9,8 +9,8 @@
 
 namespace Zend\Stratigility;
 
-use Psr\Http\Message\ServerRequestInterface as Request;
-use Psr\Http\Message\ResponseInterface as Response;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\ResponseInterface;
 
 /**
  * Middleware.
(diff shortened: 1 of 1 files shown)


weierophinney:
This interface is going away in version 2.0, which will happen shortly after the upcoming 1.3 release. As such, I will leave it as-is for now, as there's no benefit to making a change at this time.

fabiocarneiro:
@weierophinney This does not break BC. What If I want to keep using 1.3 forever?

IMO this makes sense no matter we being close to the next major or not.

weierophinney:
Implementations *do not need to use the same aliases*. All they need to do is ensure the typehints resolve to the same types as defined in the interface. There is no tangible benefit to either making the change or keeping it as-is, and in such situations, staying with what is already written is generally preferred over making a change with zero real-world impact.

PR #75, opened 2016-11-02
Provide http-interop compatibility

weierophinney:
This patch provides compatibility with http-interop 0.2.0, via the following features:

- `MiddlewarePipe`:
  - now implements `ServerMiddlewareInterface`.
  - now allows piping http-interop `ServerMiddlewareInterface` and `MiddlewareInterface` instances.
  - now allows piping callable middleware with the signature:

    ```php
    function ($request, DelegateInterface $delegate) : ResponseInterface
    ```

  - Now exposes a `setResponsePrototype()` method; when set, callable middleware implementing the legacy interface will be decorated as `ServerMiddlewareInterface` instances.

- `Next`:
  - Now implements `DelegateInterface`.

- `Dispatch`:
  - Now provides a `process()` method mimicing the `DelegateInterface`; this method is used by `Next` when dispatching `ServerMiddlewareInterface` instances.
  - Now identifies http-interop middleware and dispatches them properly.

Additionally, this patch provides decorator classes for decorating legacy callable middleware, callable middleware compatible with `ServerMiddlewareInterface`, and callable delegates. This latter feature can be used to adapt existing callable middleware classes to implement `ServerMiddlewareInterface`:

```php
use Psr\Http\Message\ResponseInterface;
use Psr\Http\Message\ServerRequestInterface;
use Psr\Http\Middleware\DelegateInterface;
use Psr\Http\Middleware\ServerMiddlewareInterface;
use Zend\Diactoros\Response\JsonResponse;
use Zend\Stratigility\Delegate\CallableDelegateDecorator;

class PingMiddleware implements ServerMiddlewareInterface
{
    public function __invoke(
        ServerRequestInterface $request,
        ResponseInterface $response,
        callable $next
    ) {
        return $this->process($request, new CallableDelegateDecorator($next, $response));
    }

    public function process(ServerRequestInterface $request, DelegateInterface $delegate)
    {
        return new JsonResponse(['ack' => time()]);
    }
}
```

Finally, the patch provides migration documentation to both v1.3 and v2.0. Another patch will be submitted shortly against the dev-2.0.0 branch that incorporates the patches from this branch, with updates to provide the final 2.0.0 functionality around http-interop compatibility.


weierophinney:
Pinging @zendframework/community-review-team , @michaelmoussa , and @xtreamwayz for review, please.

doc/book/api.md
```diff
@@ -83,11 +138,44 @@ return the response provided to you.
 > raise a deprecation notice during runtime. We will be removing the argument
 > entirely starting with version 2.0.0.
 
-As examples:
+> ### $response argument
+>
+> Using the `$response` argument is unsafe when using delegation, as an inner
+> layer could return an entirely different response, ignoring any changes you
+> may have introduced previously. Additionally, hen manipulating the response
```
michaelmoussa: ITYM "Additionally, when" unless poultry. :)

src/Dispatch.php
```diff
@@ -57,37 +65,214 @@ public function __invoke(
...
+     * @param RequestInterface $request
+     * @return bool
+     * @throws Exception\MissingResponsePrototypeException if non-interop
+     *     middleware is detected, but no response prototype is available.
+     * @throws Exception\InvalidRequestTypeException if non-interop middleware
+     *     is detected, but the request provided is not a server-side request.
+     */
+    private function isNotInteropMiddleware($handler, RequestInterface $request)
+    {
+        if ($handler instanceof ServerMiddlewareInterface || $handler instanceof InteropMiddlewareInterface) {
```
michaelmoussa: Considered this for readability?

```php
if ($this->isInteropMiddleware($handler) {
    return false;
}
```

src/Dispatch.php
```diff
@@ -57,37 +65,214 @@ public function __invoke(
...
+        if (! $request instanceof ServerRequestInterface) {
+            throw new Exception\InvalidRequestTypeException(
+                'Caught Throwable from http-interop middleware, but unable to handle '
+                . 'because request is not a server request',
+                $throwable->getCode(),
+                $throwable
+            );
+        }
+
+        printf("[%s] %s (%d)\n", get_class($throwable), $throwable->getMessage(), $throwable->getCode());
```
michaelmoussa: Would it not be better to record this as a user warning / error (perhaps using `error_log(...)` rather than outputting it? Seems like an unlikely case, but in the event that it's hit, it could leak something sensitive as as plain output.

src/Dispatch.php
weierophinney: Agreed - nice suggestion!

src/Dispatch.php
michaelmoussa: Sounds like an invitation to bring the 100% test coverage hammer down in this repository too, muahahahaha! :)

doc/book/migration/to-v2.md
```diff
@@ -131,6 +131,353 @@ work in all version 1 releases as well.
...
+pipeline is invoked via its `__invoke()` method, it will set the prototype from
+the provided `$response` argument. When present, any callable, non-error
+middleware piped to the pipeline will be wrapped in a
+`Zend\Stratigility\Middleware\CallableMiddlewareWrapper` instance, which
+converts it into an http-interop middleware type; when processed, the response
+prototype will be passed to the callable for the response argument.
+
+Starting in version 2.0.0, `MiddlewarePipe` *will no longer implement
+`Zend\Stratigility\MiddlewareInterface`, and only implement the http-interop
+`ServerMiddlewareInterface`*. This has several repurcussions.
```
xtreamwayz: repercussions ?

src/Delegate/CallableDelegateDecorator.php
```diff
@@ -0,0 +1,50 @@
...
+    public function __construct(callable $delegate, ResponseInterface $response)
+    {
+        $this->delegate = $delegate;
+        $this->response = $response;
+    }
+
+    /**
+     * Proxies to the underlying callable delegate to process a request.
+     *
+     * {@inheritDocs}
```
xtreamwayz: Should be `{@inheritDoc}`.

src/Dispatch.php
```diff
@@ -57,37 +65,213 @@ public function __invoke(
         ResponseInterface $response,
         callable $next
     ) {
-        $handler  = $route->handler;
+        if (! $this->responsePrototype) {
```
xtreamwayz: Missing method level docblock:

```
@return ResponseInterface
@throws Exception\InvalidRequestTypeException
@throws Exception\MissingResponsePrototypeException
```

src/Middleware/CallableInteropMiddlewareWrapper.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $this->middleware = $middleware;
+    }
+
+    /**
+     * {@inheritDocs}
+     */
+    public function process(ServerRequestInterface $request, DelegateInterface $delegate)
+    {
+        return $middleware($request, $delegate);
```
xtreamwayz: `return $this->middleware($request, $delegate);`

Probably lacks testing as well.

src/Middleware/ErrorHandler.php
```diff
@@ -9,13 +9,15 @@
...
+use Interop\Http\Middleware\ServerMiddlewareInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Throwable;
 use Zend\Escaper\Escaper;
+use Zend\Stratigility\Delegate\CallableDelegateDecorator;
 use Zend\Stratigility\Exception\MissingDelegateException;
 use Zend\Stratigility\Exception\MissingResponseException;
-use Zend\Stratigility\MiddlewareInterface;
 use Zend\Stratigility\Utils;
```
xtreamwayz: Unused imports

```
use Zend\Escaper\Escaper;
use Zend\Stratigility\Exception\MissingDelegateException;
use Zend\Stratigility\Utils;
```

src/Route.php
```diff
@@ -35,14 +37,24 @@ class Route
 
     /**
      * @param string $path
-     * @param callable $handler
+     * @param callable|InteropMiddlewareInterface|ServerMiddlewareInterface $handler
      */
-    public function __construct($path, callable $handler)
+    public function __construct($path, $handler)
```
xtreamwayz: I notice an inconsistent use of `@throws` in dockblocks so I'm not sure if they are required or not.
e.g. in this class the constructor lacks `@throws` in its dockblock whereas `__get()` has it.

src/Dispatch.php
weierophinney: In this particular case, neither of those two exceptions can occur:

- Since were passing the `$request` provided to `__invoke()`, `InvalidRequestTypeException` will not be raised, as we have a server-side request.
- Since we received a response, the response prototype will be present, so `MissingResponseProtototypeException` will not be thrown, either.

I *will* add the missing `@return` annotation, however!

weierophinney:
@xtreamwayz I've addressed all your points of feedback; thanks! Additionally, wrote tests for the `CallableMiddlewareWrapper` as well as the `CallableInteropMiddlewareWrapper` to ensure behavior is as expected.

xtreamwayz:
LGTM.

src/Dispatch.php
DASPRiD: Also, @throws only makes sense when the method itself throws that exception, not some dependency.

Issue #77, opened 2016-11-07
Issues with error handling in upcoming 1.3


weierophinney:
I'm preparing an RFC for error handling in Expressive 1.1, and discovered some potential problems with how error handling works for the upcoming 1.3 release of Stratigility.

While the `ErrorHandler` class is fantastic, it provides one fundamental problem: the try/catch block will have no effect out-of-the-box currently.

The reason is because `Dispatch` already has a try/catch block, and, when a Throwable or Exception is caught, it calls on the next *error* middleware, passing the Throwable or Exception:

```php
return $next($request, $response, $exception);
```

In turn, this will now raise a deprecation notice, as the third argument to `$next()` is deprecated.

To make this work, I see the following potential paths.

### Flag

First, we could introduce a flag for `MiddlewarePipe`, `raiseExceptions`. When enabled, the flag will be passed to the constructor of `Next`, which will in turn pass it on to `Dispatch`. Internally, if that flag is enabled in `Dispatch`, it will either *not* wrap middleware dispatch in a try/catch block, or re-raise the caught exception. Applications can then opt-in to this behavior in the 1.3 series; it would only technically be needed for the outermost middleware, though composed middleware pipelines should likely opt-in as well.

This flag would be marked deprecated from the outset, and removed for version 2.0.

Problems with the approach:

- Requires intervention in order to enable the flag.
- Depending on how the flag is enabled, there may be breakage when moving to 2.0. For example, if using a setter method, removing that method in 2.0 would lead to a BC break for users. If using a constructor argument, no issue would be present, but detecting the constructor argument and notifying users later to update and remove it would pose problems.

### Error middleware

Alternately, we could introduce a special error middleware. This could compose an `ErrorGenerator` and potentially error handler listeners, and then be injected into the middleware pipeline, as the last error middleware. When invoked, it would delegate to the error generator and, if incorporating listeners, trigger them.

The problems with the approach are:

- It requires changing existing pipelines to add the middleware.
- Any triggering of the middleware will de facto result in a deprecation notice, as it is *error* middleware.
- Users will need to *remove* the middleware before upgrading to 2.0, as that version does not allow error middleware.

### Special "Final Handler"

[...]

PR #78, opened 2016-11-07
Allow opt-in to raising exceptions instead of error middleware

weierophinney:
This patch implements the features suggested in #77. Specifically, it adds a "raise throwables" flag that may be set on each of the `MiddlewarePipe`, `Next`, and `Dispatch` instances; setting it on one enables it on each layer deeper:

- Setting it on `Dispatch` only enables it on that instance.
- Setting it on `Next` also enables it on `Dispatch`.
- Setting it on `MiddlewarePipe` also enables it on `Next` and `Dispatch`.

The flag is toggled off by default (current 1.0 behavior), and cannot be disabled once enabled.

Once enabled:

- `Dispatch` will not wrap middleware dispatch in a try/catch block.
- `Next`, when presented with an `$err` argument will:
  - throw the argument, if throwable.
  - raise a deprecation notice otherwise, and then create a `Zend\Stratigility\Exception\MiddlewareException` based on the `$err` value.
- `MiddlewarePipe` will enable the flag on `Next` instances it creates (and thus the `Dispatch` instances).

The end result is that, if enabled on the outermost application pipeline, you will now have the 2.0 error handling mechanism in place: middleware raises exceptions, and outer middleware catches these and does something with them.

This feature will allow Expressive, for instance, to opt-in to the feature in the next minor version, while providing default error handling mimicing what it already provides via the `FinalHandler` implementation, *without breaking backwards compatibility*.

### DIFF
--- a/src/Dispatch.php
+++ b/src/Dispatch.php
@@ -28,6 +28,14 @@ use Psr\Http\Message\ServerRequestInterface;
 class Dispatch
 {
     /**
+     * Flag indicating whether or not to raise throwables during dispatch; when
+     * false, a try/catch block is used instead (default behavior).
+     *
+     * @var bool
+     */
+    private $raiseThrowables = false;
+
+    /**
      * @var ResponseInterface
      */
     private $responsePrototype;
@@ -101,6 +109,17 @@ class Dispatch
         }
 
         return $this->dispatchInteropMiddleware($route->handler, $next, $request);
+    }
+
+    /**
+     * Enables the "raise throwables", causing this instance to raise
+     * throwables instead of catch them.
+     *
+     * @return void
+     */
+    public function raiseThrowables()
+    {
+        $this->raiseThrowables = true;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/Dispatch.php
```diff
@@ -28,6 +28,14 @@
 class Dispatch
```
xtreamwayz: Import `Interop\Http\Middleware\DelegateInterface` is not used. However it is mentioned as a comment in the docblock right before the process method, but I don't think that's a reason to import it.

src/Dispatch.php
```diff
@@ -193,6 +212,18 @@ private function dispatchCallableMiddleware(
                 break;
         }
 
+        if ($this->raiseThrowables) {
+            if ($hasError && $arity === 4) {
+                return $middleware($err, $request, $response, $next);
```
xtreamwayz: This is not tested. Not sure if it's needed though.

src/Dispatch.php
```diff
@@ -193,6 +212,18 @@ private function dispatchCallableMiddleware(
...
+            }
+
+            if (! $hasError && $arity < 4) {
+                return $middleware($request, $response, $next);
+            }
+
+            return $next($request, $response, $err);
```
xtreamwayz: Tests are missing for this.

weierophinney: Tests have now been added for both this path, and the one noted above.

src/Dispatch.php
weierophinney: Thanks, removed now.

PR #86, opened 2016-11-11
Hotfix - deprecated messages - link to migration guide

webimpress:
There are wrong links to migration guide in deprecated error messages.
There should be:
`https://docs.zendframework.com/zend-stratigility/migration/to-v2/#original-request-response-and-uri`
instead of:
`https://docs.zendframework.com/migration/to-v2/#original-request-response-and-uri`

(so there is missing `/zend-stratigility` in the path).


weierophinney:
Thanks, @webimpress

PR #89, opened 2016-11-27
Fixed typos in CONTRIBUTING.md

schnittstabil:
Fixed typos in `CONTRIBUTING.md`
- Added composer.phar to `.gitignore` to better support the suggested workflow

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,6 @@ zf-mkdoc-theme/
 
 clover.xml
 composer.lock
+composer.phar
 coveralls-upload.json
 phpunit.xml
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ To run tests:
 
   ```console
   $ git clone git@github.com:zendframework/zend-stratigility.git
-  $ cd
+  $ cd zend-stratigility
   ```
 
 - Install dependencies via composer:
@@ -66,7 +66,7 @@ To run tests:
 - Run the tests using the "test" command shipped in the `composer.json`:
 
   ```console
-  $ composer test
+  $ ./composer.phar test
   ```
 
 You can turn on conditional tests with the `phpunit.xml` file.
@@ -84,13 +84,13 @@ section on running tests.
 To run CS checks only:
 
 ```console
-$ composer cs-check
+$ ./composer.phar cs-check
 ```
 
 To attempt to automatically fix common CS issues:
 
 ```console
-$ composer cs-fix
+$ ./composer.phar cs-fix
 ```
 
 If the above fixes any CS issues, please re-run the tests to ensure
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -13,5 +13,6 @@ zf-mkdoc-theme/
 
 clover.xml
 composer.lock
+composer.phar
```
xtreamwayz: Shouldn't `composer.phar` be in `.gitignore_global`?

schnittstabil:
@xtreamwayz I don't like to have `composer.phar` in package directory either, it just is the suggested approach – I'm willing to change it to a global composer approach if the PO wants to.

However, this PR is not only about the `composer.phar` related stuff, the bugged `cd zend-stratigility` and the non-working `git clone` are fixed too.

CONTRIBUTING.md
```diff
@@ -107,7 +107,7 @@ pull your work into the master repository. We recommend using
 3. Clone the canonical repository locally and enter it.
 
    ```console
-   $ git clone git://github.com:zendframework/zend-stratigility.git
+   $ git clone git@github.com:zendframework/zend-stratigility.git
```
weierophinney: Original was correct; most developers will not have push access to the origin repository, so read-only access is fine here; it's also faster.

.gitignore
weierophinney: This is an unrelated change. It is up to the developer to ensure they have access to composer and do not pollute the project tree; we will, of course, reject commits that include the file when merging anyways. The files listed in `.gitignore` are those that could appear due to running packaging/QA/documentation tools, and are specific to the needs already in the repository.

CONTRIBUTING.md
```diff
@@ -66,7 +66,7 @@ To run tests:
 - Run the tests using the "test" command shipped in the `composer.json`:
 
   ```console
-  $ composer test
+  $ ./composer.phar test
```
weierophinney: Instead of changing these all to `composer.phar`, we should be changing `composer.phar` usage to `composer`.

CONTRIBUTING.md
schnittstabil: The original simply does not work; did you mean `git clone https://github.com/…`?

.gitignore
schnittstabil: :+1: With `s/composer.phar/composer/` this is not needed anymore.

CONTRIBUTING.md
weierophinney: No, I meant the `git://` protocol. Missed that there's a problem with it — the `:` following `github.com` should be a slash (`/`), and then it will work.

weierophinney: As in:

```console
$ git clone git://github.com/zendframework/zend-stratigility.git
```

weierophinney:
Superseded by #97.

CONTRIBUTING.md
schnittstabil: I see, but as far as I know, the `git://` protocol is not officially supported, [see for example, Github Help: Which remote URL should I use?](https://help.github.com/articles/which-remote-url-should-i-use/)

PR #91, opened 2016-11-29
Added missing tests for exception

malukenho:


test/Exception/MiddlewareExceptionTest.php
```diff
@@ -0,0 +1,45 @@
...
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-stratigility/blob/master/LICENSE.md New BSD License
+ */
+
+namespace ZendTest\Stratigility\Exception;
+
+use PHPUnit_Framework_TestCase as TestCase;
+use Zend\Stratigility\Exception\MiddlewareException;
+
+class MiddlewareExceptionTest extends TestCase
```
weierophinney: This is the only test that makes sense to me; the others can be covered by ensuring we have tests for the other classes that raise the specific exceptions. This particular one has a named constructor, which makes it a different case entirely.

weierophinney: Actually... this exception type goes away in version 2.0.0, meaning these tests are no longer necessary anyways. I'll close this issue once we've released.

weierophinney:
Closing; #97 removes `MiddlewareException`.

PR #94, opened 2016-12-16
doc: fix http-interop namespace

pine3ree:
I am not sure that was don on purpose anticipating php-fig acceptance of http-interop interfaces (now still proposed in psr)


weierophinney:
Funny, I just fixed these on develop. :smile:

weierophinney:
Thanks, @pine3ree

PR #95, opened 2017-01-05
Ensure that MiddlewarePipes are called as callables if an error is present

weierophinney:
As reported in zendframework/zend-expressive#416, error middleware nested inside a `MiddlewarePipe` was not being dispatched. This was due to the fact that `Dispatch` was identifying the pipeline as http-interop middleware, and thus dropping the `$err` argument (as interop middleware cannot accept that argument).

Theis patch updates `Dispatch` to check if the middleware is a `MiddlewarePipe` and a non-null `$err` is present; if so, it now dispatches it as callable middleware instead of as interop middleware.

### DIFF
--- a/test/DispatchTest.php
+++ b/test/DispatchTest.php
@@ -441,9 +441,6 @@ class DispatchTest extends TestCase
         );
     }
 
-    /**
-     * @todo Remove this test for version 2.0
-     */
     public function testInvokingWithMiddlewarePipeAndErrorDispatchesNextErrorMiddleware()
     {
         $error    = new RuntimeException('expected');
@@ -486,9 +483,6 @@ class DispatchTest extends TestCase
         );
     }
 
-    /**
-     * @todo Remove this test for version 2.0
-     */
     public function testInvokingWithMiddlewarePipeAndNoErrorDispatchesAsInteropMiddleware()
     {
         $request  = $this->prophesize(ServerRequestInterface::class)->reveal();
--- a/test/DispatchTest.php
+++ b/test/DispatchTest.php
@@ -442,6 +442,99 @@ class DispatchTest extends TestCase
     }
 
     /**
+     * @todo Remove this test for version 2.0
+     */
+    public function testInvokingWithMiddlewarePipeAndErrorDispatchesNextErrorMiddleware()
+    {
+        $error    = new RuntimeException('expected');
+        $request  = $this->prophesize(ServerRequestInterface::class)->reveal();
+        $response = $this->prophesize(ResponseInterface::class)->reveal();
+        $expected = $this->prophesize(ResponseInterface::class)->reveal();
+
+        $next = $this->prophesize(Next::class);
+        $next->willImplement(DelegateInterface::class);
+        $next
+            ->__invoke(
+                $request,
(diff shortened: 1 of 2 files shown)


test/DispatchTest.php
```diff
@@ -442,6 +442,99 @@ public function testInvokingWithInteropMiddlewareDispatchesIt()
     }
 
     /**
+     * @todo Remove this test for version 2.0
+     */
+    public function testInvokingWithMiddlewarePipeAndErrorDispatchesNextErrorMiddleware()
+    {
+        $error    = new RuntimeException('expected');
```
Ocramius: Consider using just `Exception`. Even better, `Throwable` may be needed

test/DispatchTest.php
```diff
@@ -442,6 +442,99 @@ public function testInvokingWithInteropMiddlewareDispatchesIt()
     }
 
     /**
+     * @todo Remove this test for version 2.0
```
Ocramius: Should probably spawn an issue for 2.0 for this, then link the issue instead

Ocramius: Also: why is this not needed in 2.0?

test/DispatchTest.php
```diff
@@ -442,6 +442,99 @@ public function testInvokingWithInteropMiddlewareDispatchesIt()
...
+        $dispatch = new Dispatch();
+
+        $this->assertSame(
+            $expected,
+            $dispatch($route, $error, $request, $this->response->reveal(), $next->reveal())
+        );
+    }
+
+    /**
+     * @todo Remove this test for version 2.0
```
Ocramius: Should probably spawn an issue for 2.0 for this, then link the issue instead

test/DispatchTest.php
weierophinney: Because this class goes away entirely, and that version will only work with http-interop middleware. :wink:

weierophinney: Which means, really, that this `@todo` can go away, as the top of the classfile already indicates the entire file goes away. Updating...

test/DispatchTest.php
weierophinney: I'll rewrite to use a data provider to seed multiple exception/throwable types. Thanks for the feedback!

test/DispatchTest.php
weierophinney: See note above; I'll just remove these, as there's a note at the top of this file already.

weierophinney:
@Ocramius Feedback incorporated; waiting for build, but should be ready for you to review.

Ocramius:
@weierophinney awesome, thanks!

Note that this cannot be forward ported, as `develop` has no `Dispatch` class anymore.

weierophinney:
> Note that this cannot be forward ported, as develop has no Dispatch class anymore.

Yep - and hence the reason I could omit the TODO items. :smile: Should not affect merging from develop to master later.

Thanks for the review; I'll get a tag out shortly.

Ocramius:
@weierophinney I'm currently tagging: want me to hold back?

weierophinney:
Nope, go ahead, @Ocramius.

PR #96, opened 2017-01-20
Update to http-middleware 0.4.1

weierophinney:
http-interop/http-middleware 0.4.1 is the latest revision, and likely the version that will be used during the FIG review period. It makes the following changes:

- The namespace changes from `Interop\Http\Middleware` to `Interop\Http\ServerMiddleware`, to indicate that the specification is only dealing with server-side middleware.
- The interface `ServerMiddlewareInterface` is renamed to `MiddlewareInterface`, as the namespace change makes it clear what type of middleware is being defined.
- The signature of `DelegateInterface::process` now accepts specifically a `ServerRequestInterface` instead of a more general `RequestInterface`.

Updating to these changes means the following breaking changes:

- We no longer support the interfaces as defined in 0.2.0.
- `Next::process()` now typehints against `ServerRequestInterface`.

The above result in a few breaking changes in the develop branch:

- `Zend\Stratigility\Next` and `Zend\Stratigility\Delegate\CallableDelegateDecorator::process` have changed signatures, due to the changes to `DelegateInterface::process`; if developers were extending these previously, their code now breaks.
- Stratigility no longer accepts middleware implementing the http-middleware 0.2.0 signatures, as those were under different namespaces and/or names.

~~I'm currently targeting this at a 1.4.0 release; however, due to the breaking changes, I could target it at a 2.0.0 release. If consensus is to take this latter route, I'll update this repository to merge in the changes currently in the develop branch as well.~~

This targets develop, which is the 2.0.0 feature branch at this time.

### DIFF
--- a/doc/book/migration/to-v2.md
+++ b/doc/book/migration/to-v2.md
@@ -157,7 +157,7 @@ To summarize:
 ### http-middleware 0.2.0 and Stratigility 1.3
 
 Starting in version 1.3.0, we offer compatibility with
-[http-interop middleware 0.2.0](https://github.com/http-interop/http-middleware/tree/ff545c87e97bf4d88f0cb7eb3e89f99aaa53d7a9).
+[http-interop/http-middleware 0.2.0](https://github.com/http-interop/http-middleware/tree/ff545c87e97bf4d88f0cb7eb3e89f99aaa53d7a9).
 That version of the specification defines the following interfaces:
 
 ```php
@@ -202,11 +202,11 @@ prototype will be passed to the callable for the response argument.
 
 ### http-middleware 0.4.1 and Stratigility 2.0
 
-http-middleware 0.4.1 introduces breaking changes in the interfaces, including
-the following:
+http-interop/http-middleware 0.4.1 introduces breaking changes in the
+interfaces, including the following:
 
 - The namespace changes from `Interop\Http\Middleware` to
-  `Iterop\Http\ServerMiddleware`, signaling a change indicating that the project
+  `Interop\Http\ServerMiddleware`, signaling a change indicating that the project
   now only targests server-side middleware.
 
 - The interface `ServerMiddlewareInterface` is now more simply
@@ -223,12 +223,13 @@ accommodate, and could have been imported in parallel to the 0.2.0 interfaces.
 However, the second represents a signature change, which has necessitated a
 major version bump in Stratigility in order to remain compatible.
 
-Stratigility 2.0.0 therefor targets http-middleware 0.4.1, and that version
-(and compatible versions) only.
+Stratigility 2.0.0 therefor targets http-interop/http-middleware 0.4.1, and that
+version (and compatible versions) only.
 
 Additionally, starting in version 2.0.0, `MiddlewarePipe` *will no longer implement
-`Zend\Stratigility\MiddlewareInterface`, and only implement the http-interop
-`MiddlewareInterface`*. This has several repercussions.
+`Zend\Stratigility\MiddlewareInterface`, and only implement the
+http-interop/http-middleware `MiddlewareInterface`*. This has several
+repercussions.
 
 ### Callable middleware in version 1.3.0
 
@@ -238,7 +239,7 @@ recommend updating your code to prepare for version 2.0.0.
 First, **we recommend *never* using the `$response` argument provided to
 middleware.**
 
-The reason for this recommendation is two-fold. First, the http-interop
+The reason for this recommendation is two-fold. First, the http-interop/http-middleware
 interfaces do not provide it, and, as such, using it within your middleware
 makes your middleware incompatible. Second, and more importantly, is due to the
 reason why http-interop does not include the argument: usage can lead to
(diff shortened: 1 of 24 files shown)


weierophinney:
Pinging @michaelmoussa and @xtreamwayz for review as well.

weierophinney:
> I'm wondering why you choose to alias Interop\Http\ServerMiddleware\MiddlewareInterface as InteropMiddlewareInterface in src. In the tests ServerMiddlewareInterface is being used.

That's a good catch.

The main reason I chose the verbiage "Interop" was to point out that it's from the http-interop project. That said, once this is ratified by php-fig, we'll be implementing both http-interop (which will be deprecated at that point) and the FIG interfaces — which do not include the "interop" verbiage. As such, I agree with you that the naming needs to change, and will push an update momentarily.

michaelmoussa:
Pretty straightforward, looks 👍 to me BUT, depending on how optimistic you are about finalization of the FIG interfaces, I'd be inclined to target 2.0.0 for this. Breaking changes are breaking changes.

If timing works out we might even manage to lump in the updating of dependencies for the other Expressive components that had major version bumps recently too.

weierophinney:
> I'd be inclined to target 2.0.0 for this. Breaking changes are breaking changes.

I hate having version inflation, but you're absolutely correct. This means, of course, that if there are any breaking changes to http-middleware between now and acceptance, we'll need a v3, but that also provides end-users with a better path forward.

I'm thinking with that, we'll also likely want to make Expressive's next version 2.0. We can provide migration tooling to help end-users adapt their existing, callable middleware into http-middleware-compatible middleware, and likely do that at the application level. (We already have *some* tools in place for this anyways.) Users would then be able to opt-in to the upgrade.

I'll get started today and merging in the other breaking changes from the dev-2.0.0 branch into this PR and/or re-target this PR against that branch.

src/Route.php
```diff
@@ -35,20 +34,17 @@ class Route
 
     /**
      * @param string $path
-     * @param InteropMiddlewareInterface|ServerMiddlewareInterface $handler
+     * @param ServerMiddlewareInterface $handler
      * @throws Exception\InvalidMiddlewareException if the $handler provided is
-     *     not an http-interop middleare type.
+     *     not an http-interop middleware type.
      */
     public function __construct($path, $handler)
```
weierophinney: I'm thinking I can use a typehint here instead of doing the checking within the body; however, doing so affects a ton of tests. I'll see how much work it is.

test/MiddlewarePipeTest.php
```diff
@@ -1,15 +1,15 @@
...
- * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2015-2017 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   https://framework.zend.com/license New BSD License
  */
 
 namespace ZendTest\Stratigility;
 
-use Interop\Http\Middleware\DelegateInterface;
-use Interop\Http\Middleware\ServerMiddlewareInterface;
-use PHPUnit\Framework\TestCase;
```
weierophinney: I wasn't aware namespacing was a thing in PHPUnit... I'll get all of these reverted.

xtreamwayz: I thought namespacing was added in version 6. But it seems there is a [forward compatibility layer](https://github.com/sebastianbergmann/phpunit/commit/ddc0e762cbfb40869135084f3088f9b3ae1627cb) since version 5.4.

test/NextTest.php
```diff
@@ -1,14 +1,14 @@
...
  * @license   https://framework.zend.com/license New BSD License
  */
 
 namespace ZendTest\Stratigility;
 
-use Interop\Http\Middleware\DelegateInterface;
-use Interop\Http\Middleware\ServerMiddlewareInterface;
+use Interop\Http\ServerMiddleware\DelegateInterface;
+use Interop\Http\ServerMiddleware\MiddlewareInterface as ServerMiddlewareInterface;
 use PHPUnit_Framework_Assert as Assert;
```
xtreamwayz: There's also a [PHPUnit\Framework\Assert](https://github.com/sebastianbergmann/phpunit/blob/5.7/src/ForwardCompatibility/Assert.php) namespace.

weierophinney: Awesome; thanks. Updated!

doc/book/migration/to-v2.md
```diff
@@ -198,9 +200,36 @@ middleware piped to the pipeline will be wrapped in a
...
-`Zend\Stratigility\MiddlewareInterface`, and only implement the http-interop
-`ServerMiddlewareInterface`*. This has several repercussions.
+### http-middleware 0.4.1 and Stratigility 2.0
+
+http-interop/http-middleware 0.4.1 introduces breaking changes in the
+interfaces, including the following:
+
+- The namespace changes from `Interop\Http\Middleware` to
+  `Interop\Http\ServerMiddleware`, signaling a change indicating that the project
+  now only targests server-side middleware.
```
bakura10: targets

Issue #101, opened 2017-03-08
Getting paths from pipeline


kynx:
I would like to be able to get a list of the paths that have been added to my `MiddlewarePipe`. My use case is this:

I am migrating an application to zend-expressive that includes several distinct sub-applications at different paths (say '/admin' and '/soap'), some of which require routing, some which don't. There is also a white label sub-application that administrators can brand and expose to their customers via a top-level path they define (ie '/mycustomer'). I need to ensure that the path they enter is not part of the core pipeline.

While I could just hard-code a list of paths it would be better if I could access something like `$app->getPipedPaths()` right after the programatic pipeline has been built. Not 100% sure if this belongs in Stratigility or Expressive - would this be of interest here?

xtreamwayz:
If you use expressive then zend-expressive would be the place to open the issue. Fortunately since Expressive 2 (released yesterday), there is [Application::getRoutes()](https://github.com/zendframework/zend-expressive/blob/2.0.0/src/Application.php#L329-L337). That might be what you are looking for.

weierophinney:
Closed, per reason stated in #103.

PR #103, opened 2017-03-09
Added `getPipedPaths()` method

kynx:
As per #101 this PR adds a method for returning the paths that have been added via `MiddlewarePipe::pipe()`

### DIFF
--- a/src/MiddlewarePipe.php
+++ b/src/MiddlewarePipe.php
@@ -188,6 +188,25 @@ class MiddlewarePipe implements ServerMiddlewareInterface
     }
 
     /**
+     * Returns array of path names that have been passed to `pipe()`
+     *
+     * @param bool $includeRoot  If true, '/' path(s) will be included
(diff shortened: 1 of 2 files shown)


weierophinney:
This is not a feature we want to support.

We deliberately do not provide mechanisms to introspect the middleware pipeline; the idea is you pipe stuff in, and the process the pipeline, making introspection an edge case.

Introspection can be accomplished in userland extensions or decorators, though you'll have to either use reflection to get access to the `$pipeline` property, or override the `pipe()` method to store the information separately.

(As an example, we already have something similar in Expressive 2.0, where we expose `getRoutes()`, which returns the set of `Route` instances that were attached to the instance, which gives access to the paths, allowed HTTP methods, and attached middleware. The example is imperfect, as routed middleware is different than pipeline middleware; the point, however, is this happens in a `MiddlewarePipe` extension.)

PR #107, opened 2017-05-15
Remove attribute raiseThrowables because Zend\Stratigility\Next is not using it

pgrau:
Remove attribute raiseThrowables because Zend\Stratigility\Next is not using it.


weierophinney:
Thanks, @pgrau

PR #112, opened 2017-10-05
Added support for http-interop/http-middleware 0.5.0 (no BC Break)

webimpress:
We keep also support for previous version. Now it depends on the library consumer what version of http-middleware to use. We accomplish that by using external library `webimpress/http-middleware-compatibility`.

Please note that tests runs with lowest/locked/latest dependencies. I've set `composer.lock` on http-middleware 0.4.1, lowest will be 0.1.1 and latest 0.5.0. All test pass :)

Probably we don't need support version http-middleware 0.1.1 anymore, so can add conflict condition in composer. To support it, I've added in two place method `next` as proxy to method `handle` (also new method, to support http-middleware 0.5.0) - and now `process` is proxy to `handle`.

Definitely we need update documentation a bit more. I'm thinking about all examples in documentation. How we should do that? Use always `http-middleware` interfaces (what version?) or aliases from `webimpress/http-middleware-compatibility`? IMHO, final library users should use http-middleware interfaces directly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,8 +22,8 @@
   "require": {
     "php": "^5.6 || ^7.0",
     "psr/http-message": "^1.0",
-    "zendframework/zend-escaper": "^2.3",
-    "http-interop/http-middleware": "^0.4.1"
+    "webimpress/http-middleware-compatibility": "^0.1.1",
+    "zendframework/zend-escaper": "^2.3"
   },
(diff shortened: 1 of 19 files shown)


webimpress:
Rebased and changed target to `develop` branch.

weierophinney:
Thanks, @webimpress!

PR #115, opened 2017-10-06
Updated PHPUnit configuration

webimpress:
- added missing namespaces
- added process uncovered files from whitelist attr
- removed add uncovered files from whitelist as it is redundant


weierophinney:
Thanks, @webimpress

PR #116, opened 2017-10-07
PSR-15 handler interface

xtreamwayz:


webimpress:
@xtreamwayz have you seen my solution in #112?
Your solution provide BC Break and we will have another one when PSR-15 become stable. I was talking with @weierophinney about my solution, and he seems to be quite happy with it right now. Of course it's temporary solution, but allows us to move forward with http-interop version without introducing BC Breaks.

xtreamwayz:
I saw your solution. I'm not sure if I like it, especially as a temporary solution. But if @weierophinney says it's ok, we go with that and this PR can be closed.

Anyway, just wanted to push the changes I made so far so they can be reviewed before I touch the docs.

webimpress:
@xtreamwayz Basically I don't think that releasing now Stratigility 3.x just because unstable http-interop middleware has been changed is a good idea. If we do like that we could finish with Stratigility 10.x with final PSR-15 😄

weierophinney:
I'm going to merge #112 for a new minor release, but keep this particular patch for when PSR-15 is ratified, and we are able to pin to that.

webimpress:
It should be closed in favour of #122

weierophinney:
Closing in favor of #122 and having an alpha followed by major release.

PR #119, opened 2017-10-11
Updated `webimpress/http-middleware-compatibility` to 0.1.3

webimpress:
This version install also composer plugin which prompt user to install `http-interop/http-middleware` as explicit dependency in their `composer.json`.
It prompts only when `http-interop/http-middleware` is not installed, otherwise it use version currently installed to update `composer.json` and message with information how to update is shown in console.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ cache:
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


weierophinney:
Thanks, @webimpress!

PR #120, opened 2017-10-12
Update docs

webimpress:
- renamed documentation directory - doc -> docs
- moved all support files to docs directory and added missing files
- updated documentation to reflect changes in Stratigility 2.1 which supports all versions of `http-interop/http-middleware`, now we have examples with different version of http-interop middleware


froschdesign:
Update the documentation is always a very good idea! 👍

weierophinney:
Thanks, @webimpress! Edited and pushed; should deploy shortly.

PR #122, opened 2017-11-13
PSR-15 - http-server-middleware [BC Break]

webimpress:
**BC Break**

Dropped `http-interop/http-middleware` support and added `http-interop/http-server-middleware`.
This package requires PHP 7.0 so I've bumped version to PHP 7.1 in this PR.

Removed also functionality marked as deprecation.


asgrim:
Probably worth linking to the discussion on this for anyone researching: https://discourse.zendframework.com/t/psr-15-compatibility-issues-and-proposal/378

docs/book/creating-middleware.md
```diff
@@ -48,42 +47,23 @@ function (
 ## http-interop middleware
 
 You can also write middleware which implements interfaces from
-`http-interop/http-middleware`. Stratigility 2.1 supports all versions of
```
weierophinney: Instead of removing these bits and pieces, I think we should likely create a "v2" (and possibly "v1") tree in the docs that details features specific to those versions, and then indicate what major version the main doc tree targets. Otherwise, we get folks still on v2 wondering why syntax they copy from the manual does not work...

docs/book/migration/to-v2.md
```diff
@@ -231,6 +231,9 @@ Additionally, starting in version 2.0.0, `MiddlewarePipe` *will no longer implem
 http-interop/http-middleware `MiddlewareInterface`*. This has several
 repercussions.
 
+Stratigility 3.0.0 targets `http-interop/http-server-middleware`, and that version
```
weierophinney: This doesn't make sense in a "migration to v2" document!

Let's create a "migration to v3" document, too.

src/Delegate/CallableDelegateDecorator.php
```diff
@@ -7,66 +7,44 @@
...
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
-use Webimpress\HttpMiddlewareCompatibility\HandlerInterface as DelegateInterface;
 
 /**
  * Decorate callable delegates as http-interop delegates in order to process
  * incoming requests.
  */
-class CallableDelegateDecorator implements DelegateInterface
+class CallableDelegateDecorator implements RequestHandlerInterface
```
weierophinney: Is this class even necessary any more? With a new major version, this might be something we do away with.

src/Middleware/CallableMiddlewareWrapper.php
```diff
@@ -44,19 +42,19 @@ public function __construct(callable $middleware, ResponseInterface $prototype)
     /**
      * Proxies to underlying middleware, using composed response prototype.
      *
-     * Also decorates the $delegator using the CallableMiddlewareWrapper.
+     * Also decorates the $handler using the CallableMiddlewareWrapper.
```
weierophinney: This sentence makes no sense. I think originally, it was saying that we wrapped the delegator in the `CallableDelegatorWrapper`. But we don't; we wrap it in a closure.

Let's remove the sentence.

src/MiddlewarePipe.php
```diff
@@ -31,7 +32,7 @@
  *
  * @see https://github.com/sencha/connect
  */
-class MiddlewarePipe implements ServerMiddlewareInterface
+class MiddlewarePipe implements MiddlewareInterface
```
weierophinney: I'm half-thinking we should have it implement `RequestHandlerInterface` as well. Doing so, hwoever, would require composing a response prototype internally so it can generate a response if none is produced by the pipeline. That could be done either via optional constructor argument, or a setter; if it is not present, and no response is produced, it would raise an exception.

We can address that at a later time, though.

src/Delegate/CallableDelegateDecorator.php
weierophinney: Never mind, I see it being used in `MiddlewarePipe::__invoke()`.

src/Next.php
```diff
@@ -67,30 +65,6 @@ public function __construct(SplQueue $queue, DelegateInterface $nextDelegate = n
      *     not return a response.
      */
     public function __invoke(ServerRequestInterface $request)
```
weierophinney: Honestly, I think we should remove this. We don't use it internally, and with the requirement that PSR-15 middleware be used, folks should be aware that they shouldn't just invoke it, but rather call the `handle()` method.

weierophinney: If you do this, though, you'll also need to update the `CallableDelegateWrapper`, as it treats `Next` instances as callables.

src/Delegate/CallableDelegateDecorator.php
danizord: @weierophinney Instead of adding internal overhead in `MiddlewarePipe` to support these wrappers, I think it would be better to provide utility functions that does the wrapping at userland level. e.g. https://github.com/danizord/mid/blob/0.1/src/mid.php#L144-L146

webimpress: @danizord I've spoke with @weierophinney and this is actually what we are doing, probably we will provide class wrappers or one class with multiple static methods. It will be done later on, probably with separate PR. Thanks for the suggestion.

weierophinney: Also, utility functions don't work well here, as we need to provide _an implementation of an interface_. Classes that accept the callback and implement the interface make sense. But we can address those in a later patch.

danizord: @weierophinney I mean an utility function that returns an anonymous object implementing the interface. It's a bit less verbose for consumers. Like that: https://github.com/danizord/mid/blob/0.1/src/mid.php#L144-L146

src/MiddlewarePipe.php
danizord: @weierophinney I'd go further and remove the response prototype. Reaching the end of middleware pipeline is actually an exceptional case and should never happen. Instead I'd recommend using a `NotFoundMiddleware` at the end of pipeline.

src/Exception/InvalidMiddlewareException.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        $received = gettype($value);
+
+        if (is_object($value)) {
+            $received = get_class($value);
+        }
+
+        return new self(
+            sprintf(
+                'Middleware must be callable, %s found',
```
weierophinney: This should change; it should indicate middleware must implement `MiddlewareInterface`.

PR #123, opened 2017-11-13
Added http-server-middleware support

webimpress:
This PR provides smooth path update from `http-middleware` to `http-server-middleware`.
Unfortunately it's very complicated... And I don't like it at all...

I think much easier will be release 3.0.0 (see PR #122) or wait for final version of PSR-15...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
   "require": {
     "php": "^5.6 || ^7.0",
     "psr/http-message": "^1.0",
-    "webimpress/http-middleware-compatibility": "^0.1.3",
+    "webimpress/http-middleware-compatibility": "dev-feature/http-server-middleware as 0.2",
     "zendframework/zend-escaper": "^2.3"
   },
   "require-dev": {
@@ -44,6 +44,9 @@
     "psr/http-message-implementation": "Please install a psr/http-message-implementation to consume Stratigility; e.g., zendframework/zend-diactoros"
   },
   "autoload": {
+    "files": [
+      "autoload/http-interop.php"
+    ],
     "psr-4": {
       "Zend\\Stratigility\\": "src/"
     }
@@ -53,6 +56,7 @@
       "ZendTest\\Stratigility\\": "test/"
     }
   },
+  "minimum-stability": "dev",
   "scripts": {
     "check": [
       "@license-check",
--- a/src/NextTrait.php
+++ b/src/NextTrait.php
@@ -19,8 +19,10 @@ use const Webimpress\HttpMiddlewareCompatibility\HANDLER_METHOD;
 
 /**
  * Iterate a queue of middlewares and execute them.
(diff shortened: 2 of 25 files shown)


composer.json
```diff
@@ -31,7 +31,7 @@
   "require": {
     "php": "^5.6 || ^7.0",
     "psr/http-message": "^1.0",
-    "webimpress/http-middleware-compatibility": "^0.1.3",
+    "webimpress/http-middleware-compatibility": "dev-feature/http-server-middleware as 0.2",
```
webimpress: I haven't released yet version 0.2, because as you can see update process will be a bit complicated and not sure if it's worth to do it. Maybe will be better to release Stratigility 3.0.0 with PHP 7.1 and http-server-middleware support and then do similar trick later on when we have accepted PSR-15. Or just wait for final PSR-15...

src/NextTrait.php
```diff
@@ -141,7 +143,11 @@ public function handle(ServerRequestInterface $request)
         }
 
         $middleware = $layer->handler;
-        $response = $middleware->process($request, $this);
+        try {
+            $response = $middleware->process($request, $this);
+        } catch (\TypeError $e) {
+            $response = null;
+        }
```
webimpress: Probably tests should be changed instead. Without this `try...catch` following tests are failing:
```
1) ZendTest\Stratigility\NextTest::testNextShouldRaiseExceptionIfMiddlewareDoesNotReturnResponse
Failed asserting that exception of type "TypeError" matches expected exception "Zend\Stratigility\Exception\MissingResponseException". Message was: "Return value of Zend\Stratigility\Middleware\CallableMiddlewareWrapperWithReturnType::process() must implement interface Psr\Http\Message\ResponseInterface, null returned" at
    .../zend-stratigility/src/Middleware/CallableMiddlewareWrapperWithReturnType.php:26
    .../zend-stratigility/src/NextTrait.php:146
    .../zend-stratigility/src/NextWithReturnType.php:25
    .../zend-stratigility/src/NextTrait.php:97
    .../zend-stratigility/src/NextTrait.php:73
    .../zend-stratigility/test/NextTest.php:235


2) ZendTest\Stratigility\NextTest::testProcessRaisesExceptionIfNoResponseReturnedByMiddleware
Failed asserting that exception of type "TypeError" matches expected exception "Zend\Stratigility\Exception\MissingResponseException". Message was: "Return value of Double\MiddlewareInterface\P90::process() must implement interface Psr\Http\Message\ResponseInterface, string returned" at
    .../zend-stratigility/src/NextTrait.php:146
    .../zend-stratigility/src/NextWithReturnType.php:25
    .../zend-stratigility/src/NextTrait.php:97
    .../zend-stratigility/test/NextTest.php:457
```

webimpress:
Closed in favour of #122

PR #124, opened 2017-11-29
Adds PSR-15 decorator classes for callable middleware

weierophinney:
This patch provides two classes, `CallableMiddlewareDecorator` and `DoublePassMiddlewareDecorator`. The former replaces the previous `CallableInteropMiddlewareWrapper`, and the latter is the new version of `CallableMiddlewareWrapper` from the v2 series.

`CallableMiddlewareDecorator` accepts a callable that is expected to conform to the PSR-15 signature basics (accepts a request and handler, and returns a response), but does no verifications on the signature. If no response is returned by the callable, it raises an exception.

`DoublePassMiddlewareDecorator` accepts a callable that follows the double-pass signature (`function ($request, $response, $next)`), but does no verifications on the signature. If no response is returned by the callable, it raises an exception.

### DIFF
--- a/src/Exception/MissingResponsePrototypeException.php
+++ b/src/Exception/MissingResponsePrototypeException.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-stratigility for the canonical source repository
+ * @copyright Copyright (c) 2017 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/zend-stratigility/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Stratigility\Exception;
+
+use UnexpectedValueException;
+use Zend\Stratigility\Middleware\DoublePassMiddlewareWrapper;
+
+/**
+ * Exception thrown by the DoublePassMiddlewareWrapper when no response
+ * prototype is provided, and Diactoros is not available to create a default.
+ */
+class MissingResponsePrototypeException extends UnexpectedValueException
+{
+    public static function create() : self
+    {
+        return new self(sprintf(
+            'Unable to create a %s instance; no response prototype provided,'
+            . ' and zendframework/zend-diactoros is not installed',
+            DoublePassMiddlewareWrapper::class
+        ));
+    }
+}
--- a/docs/book/api.md
+++ b/docs/book/api.md
@@ -173,7 +173,56 @@ your users.
 
 Stratigility provides several concrete middleware implementations.
 
-#### ErrorHandler and NotFoundHandler
+### CallableMiddlewareDecorator
+
+`Zend\Stratigility\Middleware\CallableMiddlewareDecorator` provides the ability
+to decorate PHP callables that have the same signature as or a compatible
+signature to PSR-15's `MiddlewareInterface`. This allows for one-off middleware
+creation when creating your pipeline:
+
+```php
+$pipeline->pipe(new CallableMiddlewareDecorator(function ($req, $handler) {
+    // do some work
+    $response = $next($req, $handler);
+    // do some work
+    return $response;
+});
+```
+
+### DoublePassMiddlewareDecorator
+
+`Zend\Stratigility\Middleware\DoublePassMiddlewareDecorator` provides the
+ability to decorate "double-pass", callable middleware (so-called because you
+pass the request _and_ response to the delegate) within a class implementing the
+PSR-15 `MiddlewareInterface`. This allows you to adapt existing middleware with
+the double-pass interface to work with Stratigility.
+
+```php
+$pipeline->pipe(new DoublePassMiddlewareDecorator(function ($req, $res, $next) {
+    // do some work
+    $response = $next($req, $res);
+    // do some work
+    return $response;
+});
+```
+
+`$next` is a callable that decorates Stratigility's `Next` instance; it
+ignores the response argument.
+
+The constructor takes an optional second argument, a response prototype. This
+will be used to pass to the middleware when it is executed. If no instance is
+provided, a zend-diactoros response instance is auto-wired. If you want to use
+an alternate PSR-7 `ResponseInterface` implementation, pass it when creating the
+decorator instance:
+
+```php
+$pipeline->pipe(new DoublePassMiddlewareDecorator(
+    $doublePassMiddleware,
+    $response
+));
+```
+
+### ErrorHandler and NotFoundHandler
 
 These two middleware allow you to provide handle PHP errors and exceptions, and
 404 conditions, respectively. You may read more about them in the
--- a/docs/book/migration.md
+++ b/docs/book/migration.md
@@ -6,8 +6,9 @@ and provide guidance on how to upgrade your application to be compatible.
 - [PHP support](#php-support)
 - [PSR-15](#psr-15)
 - [Pipeline (`MiddlewarePipe`)](#pipeline-middlewarepipe)
-- [Removed classes and exceptions](#removed-classes-and-exceptions)
 - [Changes in public interfaces](#changes-in-public-interfaces)
+  - [Class additions](#class-additions)
+  - [Removed classes and exceptions](#removed-classes-and-exceptions)
   - [Signature changes](#signature-changes)
   - [Removed methods](#removed-methods)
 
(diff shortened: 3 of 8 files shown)


weierophinney:
License check docheader needs updates to use `https` for all URIs in template; otherwise, tests pass.

docs/book/migration/to-v3.md
```diff
@@ -41,21 +43,55 @@ If you wish to use those types, you will need to decorate them in a
 
 The following classes have been removed:
 
-- `Zend\Stratigility\CallableDelegateDecorator`
-- `Zend\Stratigility\CallableInteropMiddlewareWrapper`
-- `Zend\Stratigility\CallableMiddlewareWrapper`
-- `Zend\Stratigility\CallableMiddlewareWrapperFactory`
+- `Zend\Stratigility\Delegate\CallableDelegateDecorator`
+- `Zend\Stratigility\Middleware\CallableInteropMiddlewareWrapper`
```
webimpress: I would add note here as well that this wrapper has been renamed to `Zend\Stratigility\Middleware\CallableMiddlewareWrapper` and "see for more details in section below"

docs/book/migration/to-v3.md
```diff
@@ -41,21 +43,55 @@ If you wish to use those types, you will need to decorate them in a
 
 The following classes have been removed:
 
-- `Zend\Stratigility\CallableDelegateDecorator`
-- `Zend\Stratigility\CallableInteropMiddlewareWrapper`
-- `Zend\Stratigility\CallableMiddlewareWrapper`
```
webimpress: The namespace here was wrong, but the class is still removed, maybe we should add note that is replaced by `Zend\Stratigility\Middleware\DoublePassMiddlewareWrapper`.

src/Exception/MissingResponsePrototypeException.php
```diff
@@ -0,0 +1,27 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-stratigility for the canonical source repository
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: 2017 or range?

src/Middleware/DoublePassMiddlewareWrapper.php
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * @throws Exception\MissingResponsePrototypeException if no response
+     *     prototype is present, and zend-diactoros is not installed.
+     */
+    public function __construct(callable $middleware, ResponseInterface $responsePrototype = null)
+    {
+        $this->middleware = $middleware;
+
+        if (! $responsePrototype && ! class_exists(Response::class)) {
+            throw Exception\MissingResponsePrototypeException::create();
```
webimpress: Not sure if I like this way of throwing exceptions, I understand this when we pass some arguments to the static method and based on these args we create some customized message, but here maybe we should use just `throw new Exception\MissingResponsePrototypeException(...)` ?

test/Middleware/DoublePassMiddlewareWrapperTest.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public function testCallableMiddlewareCanDelegateViaHandler()
+    {
+        $response = $this->prophesize(ResponseInterface::class);
+        $request  = $this->prophesize(ServerRequestInterface::class);
+
+        $handler = $this->prophesize(RequestHandlerInterface::class);
+        $handler
+            ->handle(Argument::that([$request, 'reveal']))
```
webimpress: Why not `->handle($request->reveal())` (or reveal response and request on prophesize and use here just `->handle($request)`?

test/Middleware/DoublePassMiddlewareWrapperTest.php
```diff
@@ -0,0 +1,91 @@
...
+            ->will([$response, 'reveal']);
```
webimpress: Why not `->willReturn($response->reveal())`, or revel on prophesize and here just `->willReturn($response)`.

docs/book/migration/to-v3.md
```diff
@@ -8,6 +8,8 @@ and provide guidance on how to upgrade your application to be compatible.
 - [Pipeline - `MiddlewarePipe`](#pipeline-middlewarepipe)
 - [Removed classes and exceptions](#removed-classes-and-exceptions)
```
webimpress: Maybe this section we should move into section below and create subsection?

src/Exception/MissingResponsePrototypeException.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Zend\Stratigility\Exception;
+
+use UnexpectedValueException;
+use Zend\Stratigility\Middleware\DoublePassMiddlewareWrapper;
+
+/**
+ * Exception thrown by the DoublePassMiddlewareWrapper when no response
+ * prototype is provided, and Diactoros is not available to create a default.
+ */
+class MissingResponsePrototypeException extends UnexpectedValueException
```
webimpress: Should we add `ExceptionInterface` for all exceptions in stratigility?

And here, not sure if we should extend `UnexpectedValueException` or maybe better `InvalidArgumentException`?

src/Exception/MissingResponsePrototypeException.php
webimpress: And "https" here please, so when the other PR (#125) will be merged we don't need to change it here again

webimpress:
Another suggestion, maybe we have these wrappers in separate namespace?
For example `Zend\Stratigility\Wrapper` or `Zend\Stratigility\Middleware\Wrapper`?
Maybe then we can also remove "Wrapper" suffix.

src/Middleware/DoublePassMiddlewareWrapper.php
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        $this->responsePrototype = $responsePrototype ?? new Response();
```
webimpress: As per https://github.com/zendframework/zend-problem-details/pull/7#discussion_r142523505 we should use `?:` instead :)

weierophinney:
> Not sure if you've seen @danizord comment here: #122 (comment)
> He suggested something a bit different:
> https://github.com/danizord/mid/blob/0.1/src/mid.php#L144-L146
> What do you think?

While I like I don't hate the idea, there are a few issues I have with it:

- Until PHP allows autoloading functions and constants, we're left with hacks such as the `composer.json` `autoload.files` approach. This is okay-ish, but I find it encourages grouping a bunch of functions in a single file, instead of one file per function, making it too easy to make changes to multiple functions in a single patch.
- It makes mocking more difficult. The majority of mocking tools are centered around mocking _classes_, including Prophecy (which has no facilities at all for functions). Anything that makes testing harder is something I want to avoid.
- It makes extension pretty much impossible. One case in particular where extension might be of interest is the `DoublePassMiddlewareWrapper`, where you might extend it to provide an alternative fallback response prototype.
- Debugging anonymous classes is... not fun, in my experience.

Finally, if the functions will be returning a class anyways, why have an intermediary? Why not just define a class with a constructor and create instances?

You can have a lot of similar fluency with classes, due to the ability to alias imports:

```php
use Zend\Stratigility\Middleware\CallableMiddlewareWrapper as c;
use Zend\Stratigility\Middleware\DoublePassMiddlewareWrapper as dp;

$pipeline->pipe(new c(function ($req, $handler) { /* ... */ });
$pipeline->pipe(new dp(function ($req, $res, $next) { /* ... */ });
```

And, of course, there's nothing preventing you from wrapping these in closures yourself:

```php
$callable = function (callable $middleware) {
    return new CallableMiddlewareWrapper($middleware);
};

$pipeline->pipe($callable(function($req, $handler) { /* ... */ });
```

I think the class-based approach actually gives us more flexibility in this regard.

docs/book/migration/to-v3.md
weierophinney: I think I'm actually going to rename both of the new middleware, to prevent any confusion that they are _new_ implementations. That will simplify the migration section a ton as well.

First step, though: I'm going to do a PR to version these docs. They're getting complicated currently, due to version-specific notes.

danizord:
@weierophinney points taken ;)

src/Exception/MissingResponsePrototypeException.php
weierophinney: Fixed.

src/Middleware/DoublePassMiddlewareWrapper.php
weierophinney: It simplifies the code creating the exception in this case because the exception name indicates what went wrong: missing response prototype. This allows me to ensure the message is consistent if we throw it elsewhere later, and also allows me to test the exception itself if desired.

test/Middleware/DoublePassMiddlewareWrapperTest.php
weierophinney: I was reading the internals of prophecy, and they indicate it's better to use the `Argument::that([$mock, 'reveal'])` and `will([$mock, 'reveal'])` forms, as those forms ensure the mock is fully defined at the time they are revealed. Additionally, calling `reveal()` evidently has more overhead in terms of resource usage, so delaying it is a good practice, as the operations will not execute if the prior expectations fail.

src/Exception/MissingResponsePrototypeException.php
weierophinney: Yes, but let's do that in a separate pull request.

src/Exception/MissingResponsePrototypeException.php
weierophinney: Done.

docs/book/migration/to-v3.md
weierophinney: Done.

docs/book/migration/to-v3.md
weierophinney: I've renamed the classes to use a `Decorator` suffix, which solves the "this class doesn't do what it used to" problem. As such, I've re-added the above lines.

PR #126, opened 2017-11-30
PHP 7.1 improvements

webimpress:
Removed catching Exception, we can now use only Throwable as we support PHP 7.1+

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "e6f42b2ffcf2b867aba88c550cb7369f",
+    "content-hash": "d2e8310d74608155310d69b8b4c0fcb3",
     "packages": [
         {
             "name": "http-interop/http-server-handler",
@@ -1667,16 +1667,16 @@
         },
         {
             "name": "squizlabs/php_codesniffer",
-            "version": "2.7.1",
+            "version": "2.9.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/squizlabs/PHP_CodeSniffer.git",
-                "reference": "9b324f3a1132459a7274a0ace2e1b766ba80930f"
+                "reference": "dcbed1074f8244661eecddfc2a675430d8d33f62"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/squizlabs/PHP_CodeSniffer/zipball/9b324f3a1132459a7274a0ace2e1b766ba80930f",
-                "reference": "9b324f3a1132459a7274a0ace2e1b766ba80930f",
+                "url": "https://api.github.com/repos/squizlabs/PHP_CodeSniffer/zipball/dcbed1074f8244661eecddfc2a675430d8d33f62",
+                "reference": "dcbed1074f8244661eecddfc2a675430d8d33f62",
                 "shasum": ""
             },
             "require": {
@@ -1741,7 +1741,7 @@
                 "phpcs",
                 "standards"
             ],
-            "time": "2016-11-30T04:02:31+00:00"
+            "time": "2017-05-22T02:43:20+00:00"
         },
         {
             "name": "symfony/console",
@@ -2147,7 +2147,7 @@
     "prefer-stable": false,
     "prefer-lowest": false,
     "platform": {
-        "php": "^5.6 || ^7.0"
+        "php": "^7.1"
     },
     "platform-dev": []
 }
--- a/src/Exception/InvalidMiddlewareException.php
+++ b/src/Exception/InvalidMiddlewareException.php
@@ -4,6 +4,7 @@
  * @copyright Copyright (c) 2015 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   https://github.com/zendframework/zend-stratigility/blob/master/LICENSE.md New BSD License
  */
+declare(strict_types=1);
 
 namespace Zend\Stratigility\Exception;
 
@@ -16,9 +17,8 @@ class InvalidMiddlewareException extends InvalidArgumentException
      * Create and return an InvalidArgumentException detailing the invalid middleware type.
      *
      * @param mixed $value
-     * @return InvalidArgumentException
      */
-    public static function fromValue($value)
+    public static function fromValue($value) : self
     {
         $received = is_object($value)
             ? get_class($value)
--- a/src/Middleware/ErrorHandler.php
+++ b/src/Middleware/ErrorHandler.php
@@ -4,6 +4,7 @@
  * @copyright Copyright (c) 2016-2017 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   https://github.com/zendframework/zend-stratigility/blob/master/LICENSE.md New BSD License
  */
+declare(strict_types=1);
 
 namespace Zend\Stratigility\Middleware;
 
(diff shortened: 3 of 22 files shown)


src/Exception/InvalidMiddlewareException.php
```diff
@@ -4,6 +4,7 @@
  * @copyright Copyright (c) 2015 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   https://github.com/zendframework/zend-stratigility/blob/master/LICENSE.md New BSD License
  */
+declare(strict_types=1);
```
weierophinney: Remove these throughout; we haven't yet made a decision as a project whether or not we want to support strict types yet, so having them here is inconsistent.

composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
```
weierophinney: Why is the lockfile changing?

src/Middleware/ErrorHandler.php
```diff
@@ -176,20 +164,13 @@ private function handleThrowable($e, ServerRequestInterface $request)
...
         /**
-         * @param int $errno
-         * @param string $errstr
-         * @param string $errfile
-         * @param int $errline
-         * @return void
          * @throws ErrorException if error is not within the error_reporting mask.
          */
-        return function ($errno, $errstr, $errfile, $errline) {
+        return function (int $errno, string $errstr, string $errfile, int $errline) : void {
```
weierophinney: Does PHP have a problem with this? Is it possible for PHP to send different types at all (such as nulls for the `$errfile` or `$errline` when generated from an anonymous class)?

src/Route.php
```diff
@@ -32,26 +32,17 @@ class Route
...
         $this->handler = $handler;
     }
 
     /**
-     * @param mixed $name
      * @return mixed
      * @throws OutOfRangeException for invalid properties
      */
-    public function __get($name)
+    public function __get(string $name)
```
weierophinney: Could typehinting the argument cause issues, being a PHP "magic" method type? (Just curious.)

src/Utils.php
```diff
@@ -25,13 +25,12 @@
...
-     * @param ResponseInterface $response
-     * @return int
+     * @todo: this util class is not used in stratigility, maybe it should be moved to expressive?
+     * @todo: do we need $error to be provided here? Maybe we need swap params and allow $error to be null?
      */
-    public static function getStatusCode($error, ResponseInterface $response)
+    public static function getStatusCode(Throwable $error, ResponseInterface $response) : int
     {
-        if (($error instanceof Throwable || $error instanceof Exception)
+        if (null !== $error
```
weierophinney: `$error` can never be null here, as it has a typehint that isn't nullable...

test/RouteTest.php
```diff
@@ -49,7 +50,7 @@ public function nonStringPaths()
      */
     public function testDoesNotAllowNonStringPaths($path)
```
weierophinney: We can likely remove this method entirely, as it existed pre-typehints.

src/Utils.php
webimpress: Please see my todo comment in line 28-29, I think before it was possible to pass `null` here.
You have right, my change is inconsistent, and we should decide what to do. Do we even need this class in Stratigility as it is no used?

composer.lock
webimpress: Because of 2997bc6 - Updated PHP_CodeSniffer.
Old version report errors on `declare(strict_types=1);` (requires spaces around `=`)

test/RouteTest.php
webimpress: Hm... yes and no... Please try:

```php
<?php

// declare(strict_types=1);

function test(string $x) {
    var_dump($x, gettype($x));
}

test(1);
test('str');
test(true);
```

and then uncomment line with `declare(strict_types=1);`.
This is why I've added this declaration and this is why I kept this test.

There is silent conversion between simple types, but for array, object, or null it will fail (`TypeError` exception). When we declare strict types we do not allow silent conversion, and this is want we want, IMHO.

src/Exception/InvalidMiddlewareException.php
webimpress: Please see my comment here https://github.com/zendframework/zend-stratigility/pull/126#discussion_r154879451

I've seen that @Xerkus added it also in his PR to zend-router and zend-mvc v4.

src/Route.php
webimpress: I can't see any issues with it. Documentation says the param type should be string:
http://php.net/manual/en/language.oop5.overloading.php#object.get

Only issue will be when we call this method manually (`$class->__get(1)`) with non-string value, then we get `TypeError`, what's right. I think we can keep it here.

test/MiddlewarePipeTest.php
```diff
@@ -117,7 +118,7 @@ public function testReturnsResponseReturnedByQueue()
         $this->assertSame($return, $result, var_export([
             spl_object_hash($return) => get_class($return),
             spl_object_hash($result) => get_class($result),
-        ], 1));
+        ], true));
```
webimpress: This change was needed because of strict_types declaration.

src/Middleware/ErrorHandler.php
webimpress: I've checked and all seems to be fine with typehints, see below example:
```php
<?php

declare(strict_types=1);

$c = new class {
    public function error() {
        trigger_error('hello');
    }
};

set_error_handler(function (int $errno, string $errstr, string $errfile, int $errline) {
    var_dump('Error', $errno, $errstr, $errfile, $errline);
});

$c->error();
````
https://3v4l.org/LjmEF

`void` as return type also works fine. It is possible to [return bool `false`](http://php.net/manual/en/function.set-error-handler.php#refsect1-function.set-error-handler-parameters), then we would need to change the return type. But as long as we are not doing it, it's fine.

src/Exception/InvalidMiddlewareException.php
weierophinney: Just because somebody else added to a PR doesn't mean it's accepted practice; in point of fact, I've not had a chance to review any of his patches recently. ;-)

I need to think a bit on your other comment.

src/Middleware/ErrorHandler.php
weierophinney: Okay, that's all fine, then.

src/Exception/InvalidMiddlewareException.php
webimpress: @weierophinney 
> Just because somebody else added to a PR doesn't mean it's accepted practice;

Yes, but actually I think it's good, and I just wanted to point it out, that not only I think like that ;-)

src/Utils.php
weierophinney: The reason we broke this into the static class method was due to folks wanting to re-use that functionality in their own response generators; they were having to copy-paste previously as `ErrorResponseGenerator` is marked final.

`ErrorResponseGenerator::__invoke()` receives either a `Throwable` or `Exception` as the first argument, always, so `$error` here will always be a `Throwable` under PHP 7.

As such, we can get rid of the `null !== $error` condition and keep the method.

test/MiddlewarePipeTest.php
weierophinney: That, and using a truthy value versus the expected boolean was just sloppy on my part. :)

src/Utils.php
webimpress: Ok, will change it then. Thanks for explanations. Changed in: 32b5928

src/Route.php
weierophinney: Okay.

src/Exception/InvalidMiddlewareException.php
weierophinney: Okay, the argument about loose typing on the string argument convinced me, at least for this component. We'll move forward with that.

PR #132, opened 2017-12-05
NotFoundHandler is not invokable

webimpress:
Method `process` should be used instead.


weierophinney:
Thanks, @webimpress; nice catch!

PR #135, opened 2017-12-14
Route value object without magic

snapshotpl:


weierophinney:
If we go the way of #134, we may not even need a `Route` instance at all!

Essentially, we'd only be composing middleware instances in the `MiddlewarePipe`, as path-based middleware would be handled by a special decorator.

Let's see how #134 shakes out, and then decide whether or not to proceed on this one.

weierophinney:
Closing, as #134 removes the class entirely, and will be merged soon.

PR #139, opened 2018-01-15
Segregate all documentation, including current version.

weierophinney:
If we segregate by version from the beginning, we can ensure that URIs do not change once the current version is replaced by a newer one.

Since `mkdocs.yml` allows us to define our own navigation, users will see the current version as the top-level navigation items, with later versions under their own trees.

This patch pushes the v3 docs under a `docs/book/v3/` tree, and then updates all internal links so they work correctly. The `mkdocs.yml` is then updated to reflect the changes to the navigation, using `v3/` urls for the top-level navigation items, and `v{X}/` URLs for each of the v1 an v2 releases.


froschdesign:
@weierophinney 
> This patch pushes the v3 docs under a docs/book/v3/ tree…

Should we use this process in all components in the future? (Your solution is simple and I would try to customize the documentation navigation for the different version / paths.)

froschdesign:
Here is a screenshot with the new template for the docs: 

zend-stratigility

https://raw.githubusercontent.com/froschdesign/zend-documentation-showcase/master/zend-stratigility.png

weierophinney:
> Should we use this process in all components in the future? (Your solution is simple and I would try to customize the documentation navigation for the different version / paths.)

I'm thinking we should.

The one problem with this approach currently is that for existing repositories, there's no way to create redirects for top-level items to the versioned items, which means once we make the switch, existing search results will lead to 404s.

On the flip side, however, it's dead-simple, and gives us versioned docs immediately.

PR #142, opened 2018-01-19
Host Middleware Decorator

webimpress:
Inspired by https://github.com/zendframework/zend-expressive/pull/370 and `PathMiddlewareDecorator`


weierophinney:
Thanks, @webimpress!

PR #143, opened 2018-01-19
Release 3.0.0 - fixes

webimpress:
- fixes in the docs (correct code opening with language)
- fixes function usage in function comments
- fixes typos in docs
- updates templates (LICENSE, ISSUE, PR, composer.json)
- drops PHPUnit 5.7 (we can use only PHPUnit 6.5+)
- fixes links in the docs
- fixes white characters at the end of the line in the docs
- optimizes imports (removes unused and alphabetical order)
- fixes test https://github.com/zendframework/zend-stratigility/pull/143/commits/94ab831a9af13f89dc5a852e894cd2fe0e4140c4
- adds functions test (`path`, `middleware`, `doublePassMidddleware`)
- adds more tests to improve coverage
- removes conditions which can't never occur


src/Exception/PathOutOfSyncException.php
```diff
@@ -1,24 +0,0 @@
...
- * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (https://www.zend.com)
- * @license   https://github.com/zendframework/zend-stratigility/blob/master/LICENSE.md New BSD License
- */
-declare(strict_types=1);
-
-namespace Zend\Stratigility\Exception;
-
-use RuntimeException;
-
-class PathOutOfSyncException extends RuntimeException implements ExceptionInterface
```
webimpress: it is no longer used, please see 97c4053

src/Middleware/PathMiddlewareDecorator.php
```diff
@@ -89,19 +89,8 @@ private function getTruncatedPath(string $segment, string $path) : string
...
-            return substr($path, $length);
-        }
-
-        if ('/' === substr($segment, -1)) {
-            // Re-try by submitting with / stripped from end of segment
-            return $this->getTruncatedPath(rtrim($segment, '/'), $path);
-        }
-
-        // Segment is longer than path; this is a problem.
-        throw Exception\PathOutOfSyncException::forPath($this->prefix, $path);
```
webimpress: It was not possible to get there, see: 97c4053

src/Middleware/PathMiddlewareDecorator.php
```diff
@@ -89,19 +89,8 @@ private function getTruncatedPath(string $segment, string $path) : string
             return '';
         }
 
-        $length = strlen($segment);
-        if (strlen($path) > $length) {
-            // Strip decorated path from start of current path
-            return substr($path, $length);
-        }
-
-        if ('/' === substr($segment, -1)) {
```
webimpress: This condition can't be satisfied, because we normalize $segment (prefix) in constructor and it doesn't have trailing `/`.

src/Middleware/PathMiddlewareDecorator.php
```diff
@@ -89,19 +89,8 @@ private function getTruncatedPath(string $segment, string $path) : string
             return '';
         }
 
-        $length = strlen($segment);
-        if (strlen($path) > $length) {
```
webimpress: This condition is redundant, always `true`, because we have opposite check in `process` method and return in case of true.

src/Middleware/PathMiddlewareDecorator.php
```diff
@@ -49,8 +49,8 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
             return $handler->handle($request);
         }
 
-        // Trim off the part of the url that matches the prefix if it is non-empty
-        $requestToProcess = (! empty($this->prefix) && $this->prefix !== '/')
```
webimpress: `$prefix` can't be empty, because we normalize it in constructor, and it is always at least `/`

test/Middleware/DoublePassMiddlewareDecoratorExceptionTest.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        $middleware = function ($request, $response, $next) {
+            return $response;
+        };
+
+        $this->expectException(MissingResponsePrototypeException::class);
+        $this->expectExceptionMessage(
+            'no response prototype provided, and zendframework/zend-diactoros is not installed'
+        );
+        include_once __DIR__ . '/../../src/Middleware/DoublePassMiddlewareDecorator.php';
```
webimpress: We need it here, because we disable autoload in setUp to not load Response (zend-diactoros is installed in dev mode).
We can test it another way - replace `class_exists` method in the namespace, but when we decide to import internal functions it's not gonna work, so this is why I've decided to write test this way - with disabling autoloader.

test/Middleware/PathMiddlewareDecoratorTest.php
```diff
@@ -141,98 +166,138 @@ public function nestedPathCombinations()
             'empty-bare-bareplus'        => ['',       'foo',    '/foobar',       false],
             'empty-bare-tail'            => ['',       'foo',    '/foo/',         true],
             'empty-bare-tailplus'        => ['',       'foo',    '/foo/bar',      true],
+            'empty-bare-taildot'         => ['',       'foo',    '/foo.bar',      true],
```
webimpress: Basically we need to check all these cases with `.` as separator/border. Not sure if it should match in case of `.`. Is it somewhere documented? Can't see it anywhere, tbh.

test/Middleware/PathMiddlewareDecoratorTest.php
```diff
@@ -141,98 +166,138 @@ public function nestedPathCombinations()
...
             'empty-tail-bare'            => ['',       'foo/',   '/foo',          true],
             'empty-tail-bareplus'        => ['',       'foo/',   '/foobar',       false],
             'empty-tail-tail'            => ['',       'foo/',   '/foo/',         true],
             'empty-tail-tailplus'        => ['',       'foo/',   '/foo/bar',      true],
+            'empty-tail-taildot'         => ['',       'foo/',   '/foo.bar',      true],
```
weierophinney: hmmm... I'm not sure this should match, because the prefix had a trailing slash, but the URI has a `\.` at that position instead, which would be indicative of a file, vs a directory. We may need to rethink whether or not `.` and `/` are equivalent.

webimpress: I think we should remove `.` matches at all. I've added these test to point the current behaviour.
See: https://github.com/zendframework/zend-stratigility/blob/release-3.0.0/src/Middleware/PathMiddlewareDecorator.php#L48

weierophinney: Let's do that, then; it seems problematic, and it's not behavior we've really documented.

If you remove those dot-based test cases, I'll merge, as the rest looks good.

weierophinney:
@webimpress Also looks like you may need to rebase to take care of conflicts...

webimpress:
@weierophinney it's now rebased

test/Middleware/CallableMiddlewareDecoratorTest.php
```diff
@@ -8,14 +8,15 @@
 
 namespace ZendTest\Stratigility\Middleware;
 
-use Interop\Http\Server\MiddlewareInterface;
 use Interop\Http\Server\RequestHandlerInterface;
```
samsonasik: should be use `Psr\Http\Server\ RequestHandlerInterface` ?

docs/book/v2/migration.md
```diff
@@ -457,11 +457,11 @@ class PingMiddleware implements ServerMiddlewareInterface
...
-> 
+>
 > In most cases, you should be able to simply change the import statements
 > within your class file. As an example, if you have the following in your
 > Stratigility 1.3-based project:
-> 
+>
 > ```php
 > use Interop\Http\Middleware\DelegateInterface;
 > use Interop\Http\Middleware\ServerMiddlewareInterface;
```
samsonasik: Should be use `Psr` namespace ?

test/Middleware/CallableMiddlewareDecoratorTest.php
webimpress: It would be in separate PR.

docs/book/v2/migration.md
webimpress: It would be in separate PR, this contains only fixes to the current code. Please not this line has not been changed in this PR at all.

weierophinney:
Thanks, @webimpress.

PR #146, opened 2018-01-23
Added MiddlewarePipeInterface as abstraction layer for MiddlewarePipe

### LINKED ISSUE #144: [3.0]MiddlewarePipe marked as `final`
On release-3.0.0 branch `MiddlewarePipe` is marked as `final`.

I have couple concerns about it, mainly the problem could be with testing, because we are not providing abstraction for the public interface. More I've wrote on discourse:
https://discourse.zendframework.com/t/rfc-expressive-3-design-changes/438/6

As discussed with @weierophinney on slack probably we should remove `final` or provide `interface`.

/cc @snapshotpl @danizord @xtreamwayz @weierophinney

webimpress:
`MiddlewarePipe` is a `final` class, so we need to have abstraction layer to make mocking easy

Fixes #144

### DIFF
--- a/test/MiddlewarePipeTest.php
+++ b/test/MiddlewarePipeTest.php
@@ -15,6 +15,7 @@ use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
 use ReflectionClass;
+use ReflectionMethod;
 use ReflectionObject;
 use Zend\Diactoros\Response;
 use Zend\Diactoros\ServerRequest as Request;
@@ -188,7 +189,7 @@ class MiddlewarePipeTest extends TestCase
         $pipeline = new MiddlewarePipe();
 
         $r = new ReflectionObject($pipeline);
-        $methods = $r->getMethods(\ReflectionMethod::IS_PUBLIC);
+        $methods = $r->getMethods(ReflectionMethod::IS_PUBLIC);
         $actual = [];
         foreach ($methods as $method) {
             if (strpos($method->getName(), '__') !== 0) {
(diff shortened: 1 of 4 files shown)


webimpress:
@weierophinney - I've added unit test into this PR, please have a look

test/MiddlewarePipeTest.php
```diff
@@ -179,4 +182,31 @@ public function testHandleProcessesEnqueuedMiddleware()
 
         $this->assertSame($response, $pipeline->handle($this->request));
     }
+
+    public function testMiddlewarePipeIsInstanceOfMiddlewarePipeInterface()
+    {
+        $pipeline = new MiddlewarePipe();
+
+        $r = new ReflectionObject($pipeline);
+        $methods = $r->getMethods(\ReflectionMethod::IS_PUBLIC);
```
weierophinney: Import `ReflectionMethod`, please.

test/MiddlewarePipeTest.php
```diff
@@ -179,4 +182,31 @@ public function testHandleProcessesEnqueuedMiddleware()
 
         $this->assertSame($response, $pipeline->handle($this->request));
     }
+
+    public function testMiddlewarePipeIsInstanceOfMiddlewarePipeInterface()
```
weierophinney: I'd rename this to `testMiddlewarePipeOnlyImplementsMiddlewarePipeInterfaceApi()`, as that's what it's really doing.

Honestly, I'm not 100% certain we need this to be as thorough as it is; checking that `MiddlewarePipe instanceof MiddlewarePipeInterface` is likely enough for our purposes. There may be reasons to add additional public methods down the line, primarily to provide optional collaborators, that would invalidate this test otherwise.

webimpress: @weierophinney 
> There may be reasons to add additional public methods down the line, primarily to provide optional collaborators, that would invalidate this test otherwise.

And this was the reason why I've created that test as thorough as it is. I think any public method added into that class, should be also defined in the abstraction later. This is the whole point of that PR.
If test fails, it will show us that something is changed, and we will have chance to think again if it's really needed there.

weierophinney: I'm convinced. Thanks for the feedback!

weierophinney:
Thanks, @webimpress!

PR #151, opened 2018-02-13
Updated test without diactoros to work also in separate process

webimpress:
Before, when the test run in whole test case some classes were already loaded


weierophinney:
Thanks, @webimpress.

PR #153, opened 2018-02-22
Use response facttories in ErrorHandler and NotFoundHandler

webimpress:


src/Middleware/ErrorHandler.php
```diff
@@ -75,19 +75,22 @@
...
+     *     empty ResponseInterface instance to update and return when returning
+     *     an error response.
      * @param null|callable $responseGenerator Callback that will generate the final
      *     error response; if none is provided, ErrorResponseGenerator is used.
      */
-    public function __construct(ResponseInterface $responsePrototype, callable $responseGenerator = null)
+    public function __construct(callable $responseFactory, callable $responseGenerator = null)
     {
-        $this->responsePrototype = $responsePrototype;
+        $this->responseFactory = function () use ($responseFactory) {
```
weierophinney: Needs a return typehint if decorating it is to be useful.

src/Middleware/NotFoundHandler.php
```diff
@@ -16,26 +17,27 @@
...
-     *     update and return when returning an 404 response.
+     * @param callable $responseFactory A factory capable of returning an
+     *     empty ResponseInterface instance to update and return when returning
+     *     an 404 response.
      */
-    public function __construct(ResponseInterface $responsePrototype)
+    public function __construct(callable $responseFactory)
     {
-        $this->responsePrototype = $responsePrototype;
+        $this->responseFactory = $responseFactory;
```
weierophinney: This one should be decorated with a closure as well.

composer.json
```diff
@@ -20,7 +20,7 @@
     },
     "require": {
         "php": "^7.1",
-        "psr/http-message": "^1.0",
+        "fig/http-message-util": "^1.1.2",
```
weierophinney: I don't understand these particular changes.

- We _depend directly_ on psr/http-message; removing it doesn't make sense.
- I'm unclear what the update to http-message-util 1.1.2 has to do with the changes in this patch.

weierophinney:
Thanks, @webimpress.

PR #155, opened 2018-02-25
Make all implementations final

snapshotpl:
To consistency with other implementation and promote good practices.


weierophinney:
Thanks, @snapshotpl.

PR #156, opened 2018-02-25
Use foreach instead array_walk

snapshotpl:
I know it's micro optimization, however - it's still optimization and also improve a little readability IMO.


weierophinney:
Thanks, @snapshotpl!

weierophinney:
Thanks, @snapshotpl!

PR #157, opened 2018-02-25
Avoid strtolower host

snapshotpl:
Lowercased host is defined by psr-7 design:

```php
    /**
     * Retrieve the host component of the URI.
     *
     * If no host is present, this method MUST return an empty string.
     *
     * The value returned MUST be normalized to lowercase, per RFC 3986
     * Section 3.2.2.
     *
     * @see http://tools.ietf.org/html/rfc3986#section-3.2.2
     * @return string The URI host.
     */
    public function getHost();
```

### DIFF
--- a/src/Middleware/HostMiddlewareDecorator.php
+++ b/src/Middleware/HostMiddlewareDecorator.php
@@ -30,7 +30,7 @@ class HostMiddlewareDecorator implements MiddlewareInterface
 
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler) : ResponseInterface
     {
-        $host = strtolower($request->getUri()->getHost());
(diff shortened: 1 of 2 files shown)


Rockstar04:
I'm not intimately familiar with Diactoros, but I don't see any case conversion on the host field in code, or any tests covering the conversion. Is this possibly covering an deficiency in its implementation?

weierophinney:
@Rockstar04 That's a bug in Diactoros, then! Could you open an issue on that project for me, please, linking to this one?

weierophinney:
Thanks, @snapshotpl!

snapshotpl:
@Rockstar04 good catch!

PR #160, opened 2018-03-08
Implement `process()` method in PathRequestHandlerDecorator

weierophinney:
The 0.4.1 versions of http-interop define a `process()` method on the `DelegateInterface`; the method must be implemented for this polyfill to work across all versions of http-interop.


weierophinney:
(Discovered while attempting to create the backports/deprecations pull request for zend-expressive.)

PR #163, opened 2018-03-15
Removed PathRequestHandlerDecorator which is marked as "to be removed in 3.0"

webimpress:
I've noticed it because of failing tests on #162


weierophinney:
Thanks, @webimpress; good catch!

PR #165, opened 2018-03-30
Fix PathMiddlewareDecorator so that path changes in decorated middleware are respected

acelaya:
This pull request provides a fix for https://github.com/zendframework/zend-expressive/issues/606.

It fixes it for stratigility 3.0, but the same bug is present in stratigility 2.2. However, porting the changes should be easy.


weierophinney:
Thanks, @acelaya! I will push a patch for the v2.2 series shortly as well, based on this.

acelaya:
Perfect!

PR #172, opened 2018-05-08
change strtolower(substr) and substr with stripos and strpos when possible

samsonasik:
- [x] Is this related to quality assurance?


weierophinney:
Thanks, @samsonasik!

PR #175, opened 2018-06-11
Using safe filter input instead super globals

mhf-ir:
Just simple documentation edit for good practice using super global variables.

### DIFF
--- a/docs/book/v3/usage.md
+++ b/docs/book/v3/usage.md
@@ -15,10 +15,10 @@ require __DIR__ . '/../vendor/autoload.php';
 $app    = new MiddlewarePipe();
 $server = Server::createServer(
   [$app, 'handle'],
-  $_SERVER,
-  $_GET,
-  $_POST,
-  $_COOKIE,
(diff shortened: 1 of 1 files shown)


docs/book/v3/usage.md
```diff
@@ -15,10 +15,10 @@ require __DIR__ . '/../vendor/autoload.php';
...
 $server = Server::createServer(
   [$app, 'handle'],
-  $_SERVER,
-  $_GET,
-  $_POST,
-  $_COOKIE,
+  filter_input_array(INPUT_SERVER) ?? [],
+  filter_input_array(INPUT_GET) ?? [],
+  filter_input_array(INPUT_POST) ?? [],
+  filter_input_array(INPUT_COOKIE) ?? [],
```
weierophinney: Honestly, I'm not sure that this is any safer than just using the superglobals if no `$definition` argument is provided to `filter_input_array`; without it, it simply returns the array verbatim (or, if empty, a `null`). As such, direct usage of the superglobals here is simpler to write, and requires less overall knowledge by readers to understand.

PR #177, opened 2018-07-21
Remove condition with after call `handle` in `ErrorHandler`

Moln:
The `handle` method have a return type of `ResponseInterface`, so it was never be executed.
And there is no unit test to handler it.


weierophinney:
Thanks, @Moln!

PR #178, opened 2018-07-22
Reuse code in MiddlewarePipe from Next

snapshotpl:
I had refactor step by step (commit by commit) so you can reviewing this pr by commits.

By first step I remove duplication of `Next` and `MiddlewarePipe` - `Next` can be deprecated.
Then I replace SplQueue to array - IMO they are fasters however I didn't any benchmarks.
I last change I move `Next` logic into Handle method - for improved readability.

### DIFF
--- a/src/EmptyPipelineHandler.php
+++ b/src/EmptyPipelineHandler.php
@@ -1,0 +1,28 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-stratigility for the canonical source repository
+ * @copyright Copyright (c) 2015-2018 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/zend-stratigility/blob/master/LICENSE.md New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Zend\Stratigility;
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+final class EmptyPipelineHandler implements RequestHandlerInterface
+{
+    private $class;
+
+    public function __construct(string $class)
+    {
+        $this->class = $class;
+    }
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        throw Exception\EmptyPipelineException::forClass($this->class);
+    }
+}
--- a/src/EmptyPipelineHandler.php
+++ b/src/EmptyPipelineHandler.php
@@ -15,14 +15,15 @@ use Psr\Http\Server\RequestHandlerInterface;
 
 final class EmptyPipelineHandler implements RequestHandlerInterface
 {
-    private $class;
+    private $className;
 
-    public function __construct(string $class)
+    public function __construct(string $className)
     {
-        $this->class = $class;
+        $this->className = $className;
     }
+
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        throw Exception\EmptyPipelineException::forClass($this->class);
+        throw Exception\EmptyPipelineException::forClass($this->className);
     }
 }
--- a/src/EmptyPipelineHandler.php
+++ b/src/EmptyPipelineHandler.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * @see       https://github.com/zendframework/zend-stratigility for the canonical source repository
- * @copyright Copyright (c) 2015-2018 Zend Technologies USA Inc. (https://www.zend.com)
+ * @copyright Copyright (c) 2019 Zend Technologies USA Inc. (https://www.zend.com)
  * @license   https://github.com/zendframework/zend-stratigility/blob/master/LICENSE.md New BSD License
  */
 
@@ -15,6 +15,9 @@ use Psr\Http\Server\RequestHandlerInterface;
 
 final class EmptyPipelineHandler implements RequestHandlerInterface
 {
+    /**
+     * @var string
+     */
     private $className;
 
     public function __construct(string $className)
--- a/src/MiddlewarePipe.php
+++ b/src/MiddlewarePipe.php
@@ -69,13 +69,12 @@ final class MiddlewarePipe implements MiddlewarePipeInterface
      */
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        if ($this->pipeline->isEmpty()) {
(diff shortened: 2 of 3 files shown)


src/MiddlewarePipe.php
```diff
@@ -69,13 +66,13 @@ public function __clone()
...
-        if ($this->pipeline->isEmpty()) {
-            throw Exception\EmptyPipelineException::forClass(__CLASS__);
+        if (empty($this->pipeline)) {
+            return $this->defaultHandler->handle($request);
         }
 
-        $nextHandler = clone $this;
-        $middleware = $nextHandler->pipeline->dequeue();
-        return $middleware->process($request, $nextHandler);
+        $middleware = array_shift($this->pipeline);
```
weierophinney: This is destructive, and will prevent multiple executions of the pipeline.

Why is that important?

Because in an async runner, the pipeline would _change state_ between processes, eventually becoming empty. This is why we had the `Next` implementation separate, as that then becomes the message passed as the `$handler` argument _within the current process_, and it receives _a clone_ of the pipeline.

snapshotpl: Ok, can we add some test to prevent change on this?

snapshotpl:
@weierophinney Following your suggestions I rollback and improve my main change: reuse Next in MiddlewarePipe.

Build fail because `__clone` is not covered now (`Next` care about clone queue). We can delete this implementation now, however it can be BC.

src/EmptyPipelineHandler.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-stratigility for the canonical source repository
+ * @copyright Copyright (c) 2015-2018 Zend Technologies USA Inc. (https://www.zend.com)
```
weierophinney: This should be just 2018, as that's the year this file is added.

src/EmptyPipelineHandler.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Zend\Stratigility;
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+final class EmptyPipelineHandler implements RequestHandlerInterface
+{
+    private $className;
```
weierophinney: Please add a docblock indicating the type for this property.

src/MiddlewarePipe.php
```diff
@@ -69,13 +69,7 @@ public function __clone()
...
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        if ($this->pipeline->isEmpty()) {
-            throw Exception\EmptyPipelineException::forClass(__CLASS__);
-        }
-
-        $nextHandler = clone $this;
-        $middleware = $nextHandler->pipeline->dequeue();
-        return $middleware->process($request, $nextHandler);
+        return $this->process($request, new EmptyPipelineHandler(__CLASS__));
```
weierophinney: Really, really like this!

src/MiddlewarePipe.php
```diff
@@ -86,9 +80,7 @@ public function handle(ServerRequestInterface $request) : ResponseInterface
      */
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler) : ResponseInterface
     {
-        $next = new Next($this->pipeline, $handler);
-
-        return $next->handle($request);
+        return (new Next($this->pipeline, $handler))->handle($request);
```
weierophinney: Also really like this; one less variable reference!

src/MiddlewarePipe.php
weierophinney: Actually, this is problematic.

Previously, we detected if the pipeline were empty _before_ attempting to `process()` the request, and raised an exception in that situation.

With this change, the exception may be raised _even if there are middleware in the pipeline_, if the last middleware calls its `$handler`.

I'd argue we should have two separate and distinct handlers:

- If the pipeline is empty, use the `EmptyPipelineHandler`.
- If the pipeline is not empty, use a different handler that raises either a `MissingResponseException` or a new exception denoting "no handler or middleware generated a response"/"pipeline exhausted".

So, it would look something like this:

```php
return $this->process(
    $request,
    $this->pipeline->isEmpty() ? new EmptyPipelineHandler(__CLASS__) : new PipelineExhaustedHandler(__CLASS__)
);
```

snapshotpl: Are you sure? It works now exactly same like before, or I need some test in current master branch to see difference.

> With this change, the exception may be raised even if there are middleware in the pipeline, if the last middleware calls its $handler.

What was the previous behavior?

src/EmptyPipelineHandler.php
weierophinney: What about folks not using an IDE or browsing the code in GitHub? ;-)

Yes, please add it.

src/EmptyPipelineHandler.php
```diff
@@ -0,0 +1,29 @@
...
+
+    public function __construct(string $className)
+    {
+        $this->className = $className;
+    }
+
+    public function handle(ServerRequestInterface $request) : ResponseInterface
+    {
+        throw Exception\EmptyPipelineException::forClass($this->className);
```
snapshotpl: I have another idea. Maybe we don't need to introduce this class, I can implement `RequestHandlerInterface` in `EmptyPipelineException`. This exception is strongly connected with request handling, so keeping this together make a sense.

snapshotpl:
@weierophinney I'm waiting for your response here :)

src/EmptyPipelineHandler.php
weierophinney: I don't understand why we'd have an exception implement `RequestHandlerInterface`. Can you demonstrate the workflow, please?

src/MiddlewarePipe.php
weierophinney: I looked at this more closely today, and realized that my objections were because I mistook this as a patch against `Next`, when, in fact, it's against `MiddlewarePipe`.  What the patch does is create a new default fallback handler for `MiddlewarePipe::handle()` to pass to `process()`. Previously, we were checking for an empty pipeline and raising an exception, OR pulling the first middleware off the pipeline and dispatching it using the MiddlewarePipe as the next handler.

The new approach removes the need to `clone` the pipeline instance, and acts in the same way as previously.

Preparing to merge it for 3.1.0.

PR #186, opened 2019-06-08
Throw EmptyPipelineException if passed in continuation is invoked more than once.

alihammad-gist:
- [x] _Are you fixing a bug?_  
       Yes
  - [x] _Detail how the bug is invoked currently._ 
          By calling the handler that is passed in the middlewares more than once.

  - [x] _Detail the original, incorrect behavior._
          The handler `Next::$fallbackHandler` is invoked more than once. The first invocation
           of $handler->handle inside a middleware drains the middleware queue in Next::queue,
           subsequent invocations of $handler->handle invoke Next::$fallbackHandler.

  - [x] _Detail the new, expected behavior._
           The handler should be allowed to be invoked only once or the behavior will be inconsistent. The first middleware
           invokes the second middleware in the pipe when calling hanlder::handle first time, 
           second time it will be invoking the handler `Next::$fallbackHandler`.

  - [x] Base your feature on the `master` branch, and submit against that branch.
  - [x] _Add a regression test that demonstrates the bug, and proves the fix._
         Added the smallest failing test, and the code to pass it in NextTest.php . Didn't remove uneeded `Next::$fallbackHandler` because one of the tests checks of its existence.
  - [ ] Add a `CHANGELOG.md` entry for the fix.

### DIFF
--- a/src/Exception/MiddlewarePipeNextHandlerAlreadyCalledException.php
+++ b/src/Exception/MiddlewarePipeNextHandlerAlreadyCalledException.php
@@ -1,0 +1,21 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-stratigility for the canonical source repository
+ * @copyright Copyright (c) 2017 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/zend-stratigility/blob/master/LICENSE.md New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Zend\Stratigility\Exception;
+
+use BadMethodCallException;
+
+class MiddlewarePipeNextHandlerAlreadyCalledException extends BadMethodCallException implements ExceptionInterface
+{
+
+    public static function create(): self
+    {
+        return new self("Cannot invoke Next handler more than once");
+    }
+}
--- a/src/Exception/MiddlewarePipeNextHandlerAlreadyCalledException.php
+++ b/src/Exception/MiddlewarePipeNextHandlerAlreadyCalledException.php
@@ -9,9 +9,9 @@ declare(strict_types=1);
 
 namespace Zend\Stratigility\Exception;
 
-use BadMethodCallException;
+use DomainException;
 
-class MiddlewarePipeNextHandlerAlreadyCalledException extends BadMethodCallException implements ExceptionInterface
+class MiddlewarePipeNextHandlerAlreadyCalledException extends DomainException implements ExceptionInterface
 {
 
     public static function create(): self
--- a/src/Exception/MiddlewarePipeNextHandlerAlreadyCalledException.php
+++ b/src/Exception/MiddlewarePipeNextHandlerAlreadyCalledException.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * @see       https://github.com/zendframework/zend-stratigility for the canonical source repository
- * @copyright Copyright (c) 2017 Zend Technologies USA Inc. (https://www.zend.com)
+ * @copyright Copyright (c) 2019 Zend Technologies USA Inc. (https://www.zend.com)
  * @license   https://github.com/zendframework/zend-stratigility/blob/master/LICENSE.md New BSD License
  */
 
--- a/src/Exception/MiddlewarePipeNextHandlerAlreadyCalledException.php
+++ b/src/Exception/MiddlewarePipeNextHandlerAlreadyCalledException.php
@@ -16,6 +16,6 @@ class MiddlewarePipeNextHandlerAlreadyCalledException extends DomainException im
 
     public static function create(): self
     {
-        return new self("Cannot invoke Next handler more than once");
+        return new self('Cannot invoke pipeline handler $handler->handle() more than once');
     }
 }
--- a/src/Next.php
+++ b/src/Next.php
@@ -42,7 +42,7 @@ final class Next implements RequestHandlerInterface
         $this->queue           = clone $queue;
         $this->fallbackHandler = $fallbackHandler;
         $this->queue->push(
-            new RequestHandlerMiddleware($fallbackHandler),
+            new RequestHandlerMiddleware($fallbackHandler)
         );
     }
 
--- a/src/Next.php
+++ b/src/Next.php
@@ -9,6 +9,8 @@ declare(strict_types=1);
 
 namespace Zend\Stratigility;
 
+use Zend\Stratigility\Middleware\RequestHandlerMiddleware;
+use Zend\Stratigility\Exception\EmptyPipelineException;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\RequestHandlerInterface;
@@ -39,12 +41,15 @@ final class Next implements RequestHandlerInterface
     {
         $this->queue           = clone $queue;
         $this->fallbackHandler = $fallbackHandler;
+        $this->queue->push(
+            new RequestHandlerMiddleware($fallbackHandler),
+        );
     }
 
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
         if ($this->queue->isEmpty()) {
-            return $this->fallbackHandler->handle($request);
+            throw EmptyPipelineException::forClass(__CLASS__);
         }
 
         $middleware = $this->queue->dequeue();
--- a/src/Next.php
+++ b/src/Next.php
@@ -54,8 +54,8 @@ final class Next implements RequestHandlerInterface
         }
 
         $middleware = $this->queue->dequeue();
-        $next = clone $this;
-        $this->queue = null;
+        $next = clone $this; // deep clone is not used intentionally
+        $this->queue = null; // mark queue as processed at this nesting level
 
         return $middleware->process($request, $next);
     }
--- a/src/Next.php
+++ b/src/Next.php
@@ -9,12 +9,11 @@ declare(strict_types=1);
 
 namespace Zend\Stratigility;
 
-use Zend\Stratigility\Middleware\RequestHandlerMiddleware;
-use Zend\Stratigility\Exception\EmptyPipelineException;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\RequestHandlerInterface;
 use SplQueue;
+use Zend\Stratigility\Exception\MiddlewarePipeNextHandlerAlreadyCalledException;
 
 /**
  * Iterate a queue of middlewares and execute them.
@@ -41,19 +40,23 @@ final class Next implements RequestHandlerInterface
     {
         $this->queue           = clone $queue;
         $this->fallbackHandler = $fallbackHandler;
-        $this->queue->push(
-            new RequestHandlerMiddleware($fallbackHandler)
-        );
     }
 
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
+        if ($this->queue === null) {
+            throw MiddlewarePipeNextHandlerAlreadyCalledException::create();
+        }
+
         if ($this->queue->isEmpty()) {
-            throw EmptyPipelineException::forClass(__CLASS__);
+            $this->queue = null;
+            return $this->fallbackHandler->handle($request);
         }
 
         $middleware = $this->queue->dequeue();
+        $next = clone $this;
+        $this->queue = null;
 
-        return $middleware->process($request, $this);
+        return $middleware->process($request, $next);
     }
 }
--- a/src/Next.php
+++ b/src/Next.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * @see       https://github.com/zendframework/zend-stratigility for the canonical source repository
- * @copyright Copyright (c) 2015-2018 Zend Technologies USA Inc. (https://www.zend.com)
+ * @copyright Copyright (c) 2015-2019 Zend Technologies USA Inc. (https://www.zend.com)
  * @license   https://github.com/zendframework/zend-stratigility/blob/master/LICENSE.md New BSD License
  */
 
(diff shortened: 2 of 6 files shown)


Xerkus:
This PR can not be accepted as it breaks intended behavior of the Next handler while not addressing the actual issue of allowing multiple invocations.

The problem you experience is due to your usage. Is there a reason why you need to invoke it multiple times?
Normally middleware acts on request then potentially delegates to next handler then acts on the response returned by next handler.

Do you use this middleware outside of its intended usage as a part of middleware pipe?
https://github.com/zendframework/zend-stratigility/blob/e62d1c5006ad4359b44d7700c78ca00939143fbc/src/MiddlewarePipe.php#L81-L84

I will keep this PR open for your feedback.

alihammad-gist:
All test were passing, is there an implicit Next behavior I am ignoring? I added the test, If the test captures a valid Use case instead of an exception please regect the PR. What is the standard behavior when middleware delegates to the next handler twice? This captures the edge case when a continuation is called twice, you are right about normally middle ware delegates to the next handler (once).

https://github.com/zendframework/zend-stratigility/blob/90da95e4b18569b9dd4fa3f193c3209ca807d6df/test/NextTest.php#L219-L239

Xerkus:
Hm. Next does not provide a way to push to queue once it has been passed to constructor, so one use case I had in mind is out.
That also means that only questionable edge case behavior is affected by pushing final handler to queue. I was too fast to reject it.

Going back to original issue, we have two options here: go ahead and explicitly forbid second invocation or allow multiple invocations.

Second option is easy to achieve by passing cloned Next handler to queued middlewares. Next handler is a hot path though and it will have a performance impact as it will have to be cloned for each piped middleware.
```php
$next = clone $this;
$middleware = $next->queue->dequeue();
return $middleware->process($request, $next);

public function __clone()
{
    $this->queue = clone $this->queue;
}
```

So, my question stands:  Is there a reason why you need to invoke it multiple times?

@weierophinney I would also like your input on the two options outlined above.

alihammad-gist:
> So, my question stands: Is there a reason why you need to invoke it multiple times?

I do not want to invoke it multiple times, I rather have it throw an exception if it is invoked more than once so I am told explicitly that I have done something wrong.

Xerkus:
After discussion in slack we came to conclusion that Next handler should not be allowed to be invoked multiple times from the same context as it constitutes bad design and leads to unpredictable and/or hard to test and debug results.

src/Next.php
```diff
@@ -39,12 +41,15 @@ public function __construct(SplQueue $queue, RequestHandlerInterface $fallbackHa
...
         $this->queue           = clone $queue;
         $this->fallbackHandler = $fallbackHandler;
+        $this->queue->push(
+            new RequestHandlerMiddleware($fallbackHandler)
+        );
     }
 
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
         if ($this->queue->isEmpty()) {
```
Xerkus: Reuse queue as a flag where null means there is no queue to process, unlike empty queue where fallback handler goes into play
```suggestion
         if ($this->queue === null) {
             throw MiddlewarePipeNextHandlerAlreadyCalledException::create();
         }
         if ($this->queue->isEmpty()) {
```

src/Next.php
```diff
@@ -39,12 +41,15 @@ public function __construct(SplQueue $queue, RequestHandlerInterface $fallbackHa
     {
         $this->queue           = clone $queue;
         $this->fallbackHandler = $fallbackHandler;
+        $this->queue->push(
```
Xerkus: Please revert, see comments below

src/Next.php
```diff
@@ -39,12 +41,15 @@ public function __construct(SplQueue $queue, RequestHandlerInterface $fallbackHa
...
     }
 
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
         if ($this->queue->isEmpty()) {
-            return $this->fallbackHandler->handle($request);
+            throw EmptyPipelineException::forClass(__CLASS__);
         }
 
         $middleware = $this->queue->dequeue();
```
Xerkus: As per Matthew suggestion in slack, replace following code with 
```php
$next = clone $this; // deep clone is not used intentially
$this->queue = null; // mark queue as processed at this nesting level
return $middleware->process($request, $next);
```
That change will ensure that current instance can not be called twice if some middleware down the pipe decides not to delegate and produce response itself.

test/NextTest.php
```diff
@@ -214,4 +215,27 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
 
         $this->assertSame($response, $next->handle($this->request));
     }
+
+    public function testCallingContinuationMoreThanOnceRaisesException()
```
Xerkus: I want test to be split into several, each asserting one of:
- Next handler can not be called twice (expect exception)
- middleware in the queue is called exactly once on multiple invocations, (expect exception)
- fallbackHandler with non-empty queue is called exactly once on multiple invocations, (expect exception)
- fallbackHandler with empty queue is called exactly once on multiple invocations
- Next Handler can not be called twice, when middleware in the queue returns response without delegating further while there are still middlewares in the queue. Assert fallback is not called.

Those are separate cases that can be covered my mostly same test code. Please use verbose test names describing those cases.

test/NextTest.php
```diff
@@ -214,4 +217,122 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
...
     }
+
+    public function testNextHandlerCannotBeInvokedTwice()
+    {
+        $this->expectException(MiddlewarePipeNextHandlerAlreadyCalledException::class);
+
+        $fallbackHandler = $this->prophesize(RequestHandlerInterface::class);
+        $fallbackHandler
+            ->handle(Argument::any())
+            ->willReturn(new Response)
```
Xerkus: CS: add parentheses. It is not currently enforced by zf coding standard checker, but will be in v2 ruleset
Also apply in all other places you added or changed.

PSR-12 draft:
> When instantiating a new class, parentheses MUST always be present even when there are no arguments passed to the constructor.

test/NextTest.php
```diff
@@ -214,4 +217,122 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
...
+                return $h->handle($req);
+            }
+        });
+
+        $this->queue->push(new DelegatingMiddleware);
+        $this->queue->push($middleware);
+        $this->queue->push(new DelegatingMiddleware);
+
+        $next = new Next($this->queue, $fallbackHandler->reveal());
+        $next->handle($this->request);
```
Xerkus: Do double invocation here and not in the middleware

test/NextTest.php
```diff
@@ -214,4 +217,122 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
...
+        $fallbackHandler
+            ->handle(Argument::any())
+            ->willReturn(new Response)
+            ->shouldBeCalledTimes(1);
+
+        $middleware = (new class () implements MiddlewareInterface {
+            public function process(ServerRequestInterface $req, RequestHandlerInterface $h): ResponseInterface
+            {
+                $h->handle($req);
+                return $h->handle($req);
```
Xerkus: Move this double invocation up into test itself, remove all but one delegating middleware

test/NextTest.php
```diff
@@ -214,4 +217,122 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
...
+
+    public function testNextHandlerCannotBeInvokedTwice()
+    {
+        $this->expectException(MiddlewarePipeNextHandlerAlreadyCalledException::class);
+
+        $fallbackHandler = $this->prophesize(RequestHandlerInterface::class);
+        $fallbackHandler
+            ->handle(Argument::any())
+            ->willReturn(new Response)
+            ->shouldBeCalledTimes(1);
```
Xerkus: remove expectation here, it is not part of the test

test/NextTest.php
```diff
@@ -214,4 +217,122 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
...
+        $fallBackHandler = $this->prophesize(RequestHandlerInterface::class);
+        $fallBackHandler
+            ->handle(Argument::any())
+            ->shouldBeCalledTimes(1);
+
+        // RequestHandlerInterface passed in this middleware points to the Next::$fallbackHandler
+        $middleware = (new class () implements MiddlewareInterface {
+            public function process(ServerRequestInterface $req, RequestHandlerInterface $h): ResponseInterface
+            {
+                $h->handle($req);
```
Xerkus: Same goes here, move double invocation up, remove all but one delegation middleware

test/NextTest.php
```diff
@@ -214,4 +217,122 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
...
+    }
+
+    public function testSecondInvocationAttemptDoesNotInvokeFinalHandler()
+    {
+        $this->expectException(MiddlewarePipeNextHandlerAlreadyCalledException::class);
+
+        $fallBackHandler = $this->prophesize(RequestHandlerInterface::class);
+        $fallBackHandler
+            ->handle(Argument::any())
+            ->shouldBeCalledTimes(1);
```
Xerkus: prophecy automatically makes return value double if return type declared on method. However, explicit return value declaration makes test more readable; please add `willReturn(new Response())`

test/NextTest.php
```diff
@@ -214,4 +217,122 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
...
+    }
+
+    public function testSecondInvocationAttemptDoesNotInvokeMiddleware()
+    {
+        $this->expectException(MiddlewarePipeNextHandlerAlreadyCalledException::class);
+
+        $fallBackHandler = $this->prophesize(RequestHandlerInterface::class);
+        $fallBackHandler
+            ->handle(Argument::any())
+            ->shouldBeCalledTimes(1);
```
Xerkus: Same here, add explicit `willReturn` and remove expectation

test/NextTest.php
```diff
@@ -214,4 +217,122 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
...
+
+        // RequestHandlerInterface passed in this middleware points to the Next::$fallbackHandler
+        $middleware = (new class () implements MiddlewareInterface {
+            public function process(ServerRequestInterface $req, RequestHandlerInterface $h): ResponseInterface
+            {
+                $h->handle($req);
```
Xerkus: Same as above, move double call out.

Replace middleware with mock to assert test condition:
```php
$middleware = $this->prophesize(MiddlewareInterface::class);
$middleware
    ->process(Argument::any())
    ->will(function(ServerRequestInterface $req, RequestHandlerInterface $h) {
        return $h($req);
    })
    ->shouldBeCalledOnce();
```

test/NextTest.php
```diff
@@ -214,4 +217,122 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
...
+    }
+
+    public function testShortCircuitingMiddlewareDoesNotEnableSecondInvocation()
+    {
+        $this->expectException(MiddlewarePipeNextHandlerAlreadyCalledException::class);
+
+        $fallBackHandler = $this->prophesize(RequestHandlerInterface::class);
+        $fallBackHandler
+            ->handle(Argument::any())
+            ->shouldNotBeCalled();
```
Xerkus: Add explicit return value

test/NextTest.php
```diff
@@ -214,4 +217,122 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
...
+
+        $middleware = (new class () implements MiddlewareInterface {
+            public function process(ServerRequestInterface $req, RequestHandlerInterface $h): ResponseInterface
+            {
+                $h->handle($req);
```
Xerkus: Move double invocation up, remove all middlewares except short circuiting followed by delegating

test/NextTest.php
```diff
@@ -214,4 +217,122 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
...
+
+    public function testSecondInvocationAttemptWithEmptyQueueDoesNotInvokeFinalHandler()
+    {
+
+        $this->expectException(MiddlewarePipeNextHandlerAlreadyCalledException::class);
+
+        $fallBackHandler = $this->prophesize(RequestHandlerInterface::class);
+        $fallBackHandler
+            ->handle(Argument::any())
+            ->shouldBeCalledTimes(1);
```
Xerkus: add explicit return value

src/Exception/MiddlewarePipeNextHandlerAlreadyCalledException.php
```diff
@@ -0,0 +1,21 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-stratigility for the canonical source repository
+ * @copyright Copyright (c) 2017 Zend Technologies USA Inc. (https://www.zend.com)
```
Xerkus: In all new files update copyright year to `2019`. You can also update copyright in files you touched to `<first year>-2019`.
Or I can also do this myself on merge.

alihammad-gist: I'll push this fix as well with others you pointed out in the next push.

test/NextTest.php
```diff
@@ -214,4 +217,101 @@ public function testMiddlewareReturningResponseShortCircuitsProcess()
 
         $this->assertSame($response, $next->handle($this->request));
     }
+
+    public function testNextHandlerCannotBeInvokedTwice()
+    {
+        $this->expectException(MiddlewarePipeNextHandlerAlreadyCalledException::class);
```
weierophinney: `expectException*()` statements should always be immediately prior to the statement that will cause the exception; otherwise, it's possible that the exception is being raised earlier, which could lead to a false positive test.

(I'm currently doing edits in my local checkout of your branch, and will push a change to this shortly, so no work for you right now, @alihammad-gist !)

alihammad-gist: Noted.

weierophinney:
Thanks, @alihammad-gist!

Issue #190, opened 2019-08-13
Naming conflict for "ErrorHandler" which actually is a middleware and not a RequestHandler


boesing:
Hey there,

I am giving some trainings regarding zend-expressive, e.g. for some colleagues.
One thing, which always came back to me as a question was:

Why is the `ErrorHandler` called `ErrorHandler` like all those request handlers and not `ErrorMiddleware` which actually suits better as its a middleware per definition.

Would like to see the next major version where the `ErrorHandler` is renamed to `ErrorMiddleware`.

Any thoughts on this?

weierophinney:
We can do this by doing the following:

- Introduce a new minor version that creates the new class, and then has `ErrorHandler` extend it. The constructor can then emit an `E_USER_DEPRECATION` message indicating the new name, and how to configure it.
- The new major version would then remove it.

Please feel free to submit a PR against develop to accomplish the first step.

boesing:
I like the idea of providing that new `ErrorMiddleware`.
I am not sure if triggering a deprecation message is a good idea.
As of `zend-expressive`, the `ErrorHandler` is being automatically set to the configuration.
So after updating `zend-stratigility`, every `zend-expressive` project will receive deprecation messages.

I would rather implement just a `@deprecated` tag, so that the handler can be safely removed in v4 instead.
A migration guide should probably mention that the registered `delegators` should attach the listeners to the new middleware, as just "aliasing" wont be enough. (Some more details can be found in this comment https://github.com/zendframework/zend-mvc/pull/294#issuecomment-421050228)

I will provide a PR shortly with the changes.

weierophinney:
This repository has been closed and moved to laminas/laminas-stratigility; a new issue has been opened at https://github.com/laminas/laminas-stratigility/issues/3.

PR #191, opened 2019-08-19
Renaming ErrorHandler to ErrorHandlerMiddleware

boesing:
As already mentioned in #190, there is a naming conflict in the `ErrorHandler`.

I had some feedback of trainees which asked me, why the `ErrorHandler` ain't a `RequestHandler`, as its name says.
That made me think about the naming and I realized, its just because of the missing `Middleware` postfix.

However, I'd like to rename the `ErrorHandler` to `ErrorHandlerMiddleware` to avoid future confusions about where the difference between `RequestHandlerInterface` and `MiddlewareInterface` is.


src/Middleware/ErrorHandlerMiddleware.php
```diff
@@ -0,0 +1,202 @@
...
+ * </code>
+ *
+ * Listeners are given the error, the request responsible, and the generated
+ * error response, and can then react to them. They are best suited for
+ * logging and monitoring purposes.
+ *
+ * Listeners are attached using the attachListener() method, and triggered
+ * in the order attached.
+ *
+ * @todo Mark this error middleware as final in v4
```
vaclavvanik: This class should not be marked as final because it has methods that are not in implemented interface(s).

boesing: As I am extending that middleware in the old class, its not possible to mark it final.
Thats why it should be marked as final in v4.

Dont wanted to duplicate code.

boesing: Oh, I missread your suggestion aswell.
Even if this class has methods which are not provided by an interface, its still possible to mark it as `final`.

Its only the `ErrorHandlerMiddleware::attachListener` and as this is a so called `setter` method, I wont provide a dedicated interface for it, as you usually do not typehint against that method. But I am not the one who will make that decision, its just a suggestion. I would prefer having this as final as there should be no need to extend that class at all.

webimpress:
@boesing I think there might be some references to this class in the documentation so we need update it. Other than that LGTM 👍

weierophinney:
This repository has been closed and moved to laminas/laminas-stratigility; a new issue has been opened at https://github.com/laminas/laminas-stratigility/issues/2.

weierophinney:
This repository has been moved to laminas/laminas-stratigility. If you feel that this patch is still relevant, please re-open against that repository, and reference this issue. To re-open, we suggest the following workflow:
 - Squash all commits in your branch (`git rebase -i origin/{branch}`)
 - Make a note of all changed files (`git diff --name-only origin/{branch}...HEAD
 - Run the laminas/laminas-migration tool on the code.
 - Clone laminas/laminas-stratigility to another directory.
 - Copy the files from the second bullet point to the clone of laminas/laminas-stratigility.
 - In your clone of laminas/laminas-stratigility, commit the files, push to your fork, and open the new PR.
 We will be providing tooling via laminas/laminas-migration soon to help  automate the process.

